PR #13, opened 2019-12-27
Local time offset was lost when writing DateTimeOffset in non UTC format

KyleGobel:
For the writing part this would be a breaking change.

This PR converts the `DateTimeOffset` to universal time before converting it the INT96.

Currently parquet-dotnet appears to write everything as is, so any local offset would be lost when writing.  When reading it back in it would be read as if it was a UTC `DateTimeOffset` (which then of course gives a different time instant than what the user attempted to write).


aloneguid:
build fails only due to internal nuget feed conflict, everything else is perfect, merging.

Issue #28, opened 2020-01-19
Ability to ignore fields/properties(of business objects) during serialization to parquet file


srigumm:
**Version:** Parquet.Net

**Runtime Version:** .Net Core

**OS:** Windows/Linux/MacOSX

#### Expected behavior
Though current serialize API is very helpful to work with business objects, it would very helpful if it can provide a way to ignore few properties during serialization. Something like the below,
```csharp
public class SimpleStructure
    {
        public int Id { get; set; }
        public string FirstName { get; set; }

        [ParquetIgnore]
        public string SSIN { get; set; }
    }
```
#### Actual behavior
 No such feature exists on our current serialize API

### Steps to reproduce the behavior

1. Step 1.
2. Step 2.

PR #31, opened 2020-01-21
Enhancement #28: Add ability to ignore properties

srigumm:
- ParequetConvert serialize API should ignore properties marked with
  ignore attribute during serailization process to parquet file.

### Fixes

Issue #28 

### Description

desctiption goes here

[YES] I have included unit tests validating this fix.
[YES] I have updated markdown documentation where required.
[ YES] I understand that successful approval of my pull request requires reproducible tests as per  [Contribution Guideline](https://github.com/aloneguid/parquet-dotnet/blob/master/.github/CONTRIBUTING.md).

### DIFF
--- a/doc/serialisation.md
+++ b/doc/serialisation.md
@@ -61,6 +61,20 @@ class MyClass
 It supports the following properties:
 - **Name** overrides column name in the parquet file.
 
+### Ignoring properties while serializing
+
+You can ignore few properties from serialization process by decorating them with `ParquetIgnore` attribute.
+
+ with `[ParquetIgnore]` attribute:
+
+```csharp
+class MyClass
+{
+   [ParquetIgnore]
+   public int Id { get; set; }
+}
+```
+
 ## Limitations
 
 At the moment serialiser supports only simple first-level class *properties* (having a getter and a setter).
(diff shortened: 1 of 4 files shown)


src/Parquet/Attributes/ParquetColumnAttribute.cs
```diff
@@ -30,4 +30,12 @@ public ParquetColumnAttribute(string name)
       /// </summary>
       public string Name { get; set; }
    }
+
+   /// <summary>
+   /// Annotates a class property as marker to ignore while serializing to parquet file
+   /// </summary>
+   [AttributeUsage(AttributeTargets.Property, AllowMultiple = false)]
+   internal class ParquetIgnoreAttribute : Attribute
```
aloneguid: You'll need to make the attribute public as opposed to `internal`, otherwise it won't be visible to other projects.
Probably a good idea to put it in a separate file.

srigumm: sure.

srigumm: I refactored this code little bit, it should be good now. Thanks.

Issue #38, opened 2020-03-03
Cannot parse Decimal types


elexisvenator:
**Version:** Parquet.Net v3.6.0

**Runtime Version:** .Net Core 2.1

**OS:** Windows

#### Expected behavior

With a parquet file that has a decimal128 column, Parquet.Net should be able to read the file, parsing that column as a Decimal.

#### Actual behavior

Null reference exception occurs.

### Steps to reproduce the behavior

1. Create a parquet file with a column with the Decimal128 data type.
2. Attempt to parse the file, i was using ParquetReader.ReadAsTable()
3. Null reference exception 

#### Code snippet reproducing the behavior

```csharp
using Parquet.Data.Rows;
using Xunit;

namespace Parquet.Test
{
   public class DecimalTypeTest : TestBase
   {
      [Fact]
      public void Read_File_As_Table_With_Decimal_Column_Should_Read_File()
      {
         const int decimalColumnIndex = 4;
         Table table = ReadTestFileAsTable("test-types-with-decimal.parquet");

         Assert.Equal(1234.56m, table[0].Get<decimal>(decimalColumnIndex));
      }
   }
}
```

[test-types-with-decimal.zip](https://github.com/aloneguid/parquet-dotnet/files/4279006/test-types-with-decimal.zip)

PR #39, opened 2020-03-03
Fix parsing of Decimal types

elexisvenator:
### Fixes

Issue #38 

### Description

Decimal types require the schema to determine precision, scale and what base type the data is.
For some reason it is passed in as null. I looked at all other implementations of `ReadSingle` and
cannot see any reason why it has to be null. So I passed in the SchemaElement.


elexisvenator:
Hi @aloneguid , would it be possible to get this merged soon? This bug is a bit of a blocker for me. :)

PR #73, opened 2020-07-23
[WIP] - Make it Async

### LINKED ISSUE #35: Async Methods
Kinda picking up off of https://github.com/elastacloud/parquet-dotnet/issues/20

I see now that the thrift lib ONLY has async methods, and the older clients are being removed next release.

I'm not sure how often parquet updates it's thrift spec, or how often thrift adds new features or w/e though, so this might not be a big deal.

Curious if In the past have you done any testing to see what kind of negative performance making some of the reading/writing async has?  

Was considering helping and trying to update to the latest `netstd` thrift generation, but I'm more interested in making this faster....not slower ;)

felipepessoto:
I did a quick exercise to change the API to Async.

I think it is better to don't merge this changes. I've never used this library, Thrift or Parquet before, so we need to really test it. (the unit tests are passing, but we need to test in a real world application).
We can use it to have an idea of the effort to change it to async, since async is all or nothing.

There are 2 thing we also need to validate, how to workaround Async Constructor (maybe use private ctor and factory method) and if it is ok to return Tuples, because async method can't have out params

PS: I built a Thrift package locally using the code in https://github.com/apache/thrift/tree/master/lib/netstd

Fix #35


felipepessoto:
The usage of BinaryReader/Writer that I found are done upon MemoryStream, so I changed only the code that call Read/Writes in the input Stream

PR #83, opened 2020-10-16
Less String Allocation with a Configurable Ring Buffer.

TikiBill:
I have a parquet file (which unfortunately I cannot share) with around 2.1M rows in it. The file covers a single month of data and contains a string field for the calendar quarter (e.g. 2020Q3). Thus the value is the same for all records. However, when it gets read, (I believe) the same string gets allocated 2.1M times -- eating up memory and slowing down parsing.

This implements a simple ring buffer with a configurable size. On my system with the particular file mentioned above, it reduces parsing time from ~14 seconds down to ~8 seconds.

Thank you.

--Bill

### DIFF
--- a/src/Parquet.Test/ParquetWriterTest.cs
+++ b/src/Parquet.Test/ParquetWriterTest.cs
@@ -255,5 +255,48 @@ namespace Parquet.Test
             Assert.Equal("value2", reader.CustomMetadata["key2"]);
          }
       }
+      
+      [Fact]
+      public void String_allocation_reduction_correct_strings()
+      {
+         //write a file with a single row group
+         var str = new DataField<string>("str");
+         var ms = new MemoryStream();
+         
+         var string1 = "SomeString1";
+         var string2 = "SomeString2";
+         var string3 = "xomeString1";
+         var repeats = 100;
+         var data = new string[repeats * 3];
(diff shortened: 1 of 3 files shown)


tpeplow:
I see the issue but I think the _right_ solution is to support [dictionary encoding](https://github.com/apache/parquet-format/blob/master/Encodings.md).  That way the dictionary will be read, each unique string will be allocated, and then a reference to that string will be passed out for each value from the column.  This will give you the perf improvement but won't be limited to a small ring buffer.  It will also mean downstream consumers will benefit when reading the file as it's idiomatic parquet at this point.

I hope to have some time to look into the dictionary encoding problem and chat with @aloneguid.

aloneguid:
Hi, yes there were a few who wanted to implement dictionary encoding for read/write but I haven't heard anything back. Not sure I have time at the moment, you're welcome to raise a PR 👍

aloneguid:
We might also want to check out zero-allocation way to create strings - `https://docs.microsoft.com/en-us/dotnet/api/system.string.create?view=net-5.0`. This might work out better than ring buffer, in combination with string interning in the .net runtime itself.

tpeplow:
I might have time in a few weeks to look at dictionary encoding, if you're happy to review what I do I'm happy to have a stab at it!

Issue #84, opened 2020-10-21
Performance issues observable with a pretty wide data set


tpeplow:
**Version:** Parquet.Net v.3.7.7

**Runtime Version:** .NET Core 3.0

**OS:** Windows

#### Issue

I have a data set with ~14k columns.  On profiling my own code I noticed some opportunities for performance improvements in this library.  I've logged this issue so I can propose some changes that have brought significant performance gains to my use case, which have knock on benefits to more sane data sets I have test coverage over.

### Steps to reproduce the behavior

My dataset is the debug state of mathematical model.  Each column denotes a variable in the formula language, each row is the value is a point in time for an insurance policy.  The table is sparsely populated.  At a very high level the data set relates to insurance policy's which are projected into the future.  Policies are issued at different times and the die or lapse at different points in time.  I need to be able to analyze this debug data set at scale and it's currently stored in a proprietary data format, which doesn't natively work in any big data platforms.  I'm building a spike to translate our custom format into Parquet such that I can leverage off the shelf tools for analytics.

Due to the complexity of the end to end use case I've tested the impact of this change from tests in my code base, in the pull request I'll issue with this issue I'll show the impact of each change.  Here is the initial state:

![image](https://user-images.githubusercontent.com/278561/96673866-59b55480-131c-11eb-8de0-8f05692000aa.png)

#### Code snippet reproducing the behavior

I think a good thing to do would be to add a few common scenarios as performance benchmarks so we can codify the expected performance characteristics and I'm happy to help do that.

I didn't want to make any significant changes, my approach here has been to be as surgical as possible with the improvements.

PR #85, opened 2020-10-21
Write performance improvements

tpeplow:
### Fixes

Issue #84 

### Description

Before these changes the use case took 107,716ms to complete.  After each of these changes the time is 10,232ms.
**Note** The total time is the end-to-end test, now Parquet.net is a small part of the overhead of the conversion algorithm, where before it was the majority of the time. 

This screen shot shows the performance after all the refactoring.
image

**Reuse `ThriftSchemaTree` to reduce GC overhead.**

image

To resolve this I construct the `ThriftSchemaTree` in the `ThriftFooter` so it doesn't need to be created for each call to `GetPath`.  The impact of this change is shown here:

image

**Introduce ArrayView**

1. Remove the LINQ multiple enumerations of the data and overhead associated with allocations of LINQ Count used to allocate the new array allocated in PackDefinitions.
1. Reduce GC overhead by renting the Array underpinning the ArrayView for the new array allocated to hold the values from Pack Definitions, which is consistent with definition levels.
1. Do not add additional enumerations for computing statistics, most notably the distinct values.  The implementation of distinct values is based on the LINQ Distinct implementation, but uses the HashSet not the lightweight internal Set used by LINQ.  The benefit of this is in Net Core 2.1 the HashSet is able to be initialized with a capacity, which has a decent impact thanks to no resizing.

image

**Optimize string comparison**

There were various places where string comparisons were used (e.g. in the Distinct operation and min/max values when computing statistics), which benefit from the faster `CompareOrdinal` implementation.

**Memoize `ThriftSchemaTree.Find`**

This was called a fair amount via GetPath and the benefit is shown below:

image

**Memoize `ThriftFooter.GetLevels`**

This was also on the hot path, benefit shown below:
image

GetLevels is also faster based on a more optimal algorithm, you'll see that change in an earlty commit before I introduced the memoization.

### DIFF
--- a/src/Parquet/Data/WritableArrayView.cs
+++ b/src/Parquet/Data/WritableArrayView.cs
@@ -1,0 +1,39 @@
+﻿using System.Buffers;
+
+namespace Parquet.Data
+{
+   class WritableArrayView<T> : ArrayView
+   {
+      private static readonly ArrayPool<T> ArrayPool = ArrayPool<T>.Shared;
+      private readonly T[] _typedArray;
+      int _count;
+
+      public WritableArrayView(int length) : this(ArrayPool.Rent(length))
+      {
+      }
+
+      WritableArrayView(T[] array) : base(array)
+      {
+         _typedArray = array;
+      }
+
+      public override int Count => _count;
+
+      protected override void ReturnArray()
+      {
+         ArrayPool.Return(_typedArray);
+      }
+
+      public T this[int i]
+      {
+         set
+         {
+            {
+               _typedArray[i] = value;
+               int count = i + 1;
+               if (count > _count) _count = count;
+            }
+         }
+      }
+   }
+}
--- a/src/Parquet/File/ThriftFooter.cs
+++ b/src/Parquet/File/ThriftFooter.cs
@@ -2,7 +2,6 @@
 using System.Collections.Generic;
 using System.IO;
 using System.Linq;
-using System.Reflection;
 using Parquet.Data;
 using Parquet.File.Data;
 using Parquet.Thrift;
@@ -284,6 +283,8 @@ namespace Parquet.File
 
       class ThriftSchemaTree
       {
+         readonly Dictionary<SchemaElement, Node> _memoizedFindResults = new Dictionary<SchemaElement, Node>();
+         
          public class Node
          {
             public Thrift.SchemaElement element;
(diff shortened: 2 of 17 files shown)


src/Parquet/Data/WritableArrayView.cs
```diff
@@ -0,0 +1,39 @@
+﻿using System.Buffers;
+
+namespace Parquet.Data
+{
+   class WritableArrayView<T> : ArrayView
```
pianomanjh, line 5: have you considered having this class be Disposable, to handle the array return to the pool?  Seems more discoverable, supports the idea a user may want to write the same data to a handful of parquet files.   May also clean up the design a bit by keeping the pooling concerns in this one place, and out of the base class.

src/Parquet/File/ThriftFooter.cs
```diff
@@ -126,6 +142,8 @@ public void GetLevels(Thrift.ColumnChunk columnChunk, out int maxRepetitionLevel
                i++;
             }
          }
+         
+         _memoizedLevels.Add(comparer, Tuple.Create(maxRepetitionLevel, maxDefinitionLevel));
```
pianomanjh, line 146: are we preferring Tuple here instead of value tuples because of the additional package dependency?  Value tuples would have no heap allocations.

src/Parquet/Data/WritableArrayView.cs
tpeplow: I did think of that, but I was trying to keep the scope of this change to a minimum.  It was a fair bit more surgical to return the array where I did as it required minimal refactoring.  

I think a broader change to encapsulate the `definitionsLevels` array that's also rented would clean the design up a bit.  It's allocated along with the `ArrayView` inside `PackDefinitions` on `BasicDataTypeHandler`.  I was thinking of making `ArrayView` a proper domain concept then I could have it handle the definition levels array too.  I'm happy to do it but didn't want to change _too much_ here.

src/Parquet/File/ThriftFooter.cs
tpeplow: I agree.  I didn't want to add in the reference without discussing it with @aloneguid first.

aloneguid:
LGTM!

tpeplow:
Thanks!  I saw you pushed a new Parq nuget reference but not a 3.8 Parquet.net - any chance you could push that up?  Thanks

aloneguid:
@tpeplow sorting it out now, for some reason the build has only generated .snupkg.

aloneguid:
@tpeplow live as 3.8.1

PR #91, opened 2020-12-03
Allows Dots in field name. Fix #82

### LINKED ISSUE #82: Field name with dot (.)
This commit https://github.com/aloneguid/parquet-dotnet/commit/16b55d54bc7f096eefb9df8227b1f321d9839ee5 added a validation to disallow fields with dots.

Do we really have a reason to don't accept dots? If I comment the throw line I can read a file with dots in column names, and the only failing test is the test that validates if it throws when we use dots.

I couldn't find any spec saying dots are not allowed.

Thanks.

felipepessoto:
### Fixes

Issue #

https://github.com/aloneguid/parquet-dotnet/issues/82

### Description

desctiption goes here

- [ ] I understand that successful approval of my pull request requires reproducible tests as per [Contribution Guideline](https://github.com/aloneguid/parquet-dotnet/blob/master/.github/CONTRIBUTING.md).

### DIFF
--- a/src/Parquet.Test/SchemaTest.cs
+++ b/src/Parquet.Test/SchemaTest.cs
@@ -134,12 +134,6 @@ namespace Parquet.Test
       }
 
       [Fact]
-      public void Cannot_use_dots_in_field_names()
-      {
-         Assert.Throws<ArgumentException>(() => new DataField<int>("one.two"));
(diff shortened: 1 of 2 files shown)


aloneguid (approved):
LGTM

PR #96, opened 2021-01-11
Fix reading of plain dictionary with zero length

aloncatz:
Sometimes parquet files generated by spark contain a page full of nulls. These pages have zero size and parquet.net fails with System.IO.EndOfStreamException.

The ReadRleBitpackedHybrid function interprets zero length as an instruction to read the actual length from the stream. ReadLevels relies on this behavior, however, zero can be a valid length when reading a plain dictionary page.

I decided not to mess with ReadRleBitpackedHybrid implementation and instead check that length is zero and skip the remaining items


aloneguid (approved):
apologies for the delay.
lgtm!

PR #109, opened 2021-04-23
Add support for decimal scale and precision

bdquig:
### Fixes

default scale and precision is used for decimals

Issue #

### Description

I followed the same methodology used for DateTime


aloneguid (approved):
lgtm, thanks!

cajuncoding:
So, this was a breaking change, as my code from v3.8.1 now breaks since apparently there are No longer any defaults.

It's great to support precision, but the defaults were good for our use cases, and now with the latest version we ran into unexpected errors.

I think that the ParquetColumnAttribute or more likely, the constructor of the DecimalDataField class should allow fallback to defaults for backwards compatiblity with prior behavior.  In, addition some Docs to help explain this when using the `ParquetColumnAttribute` for name mapping would be very helpful!

Just commenting here, but I'll create a new issue to track this request for backwards compatibility.

Issue #111, opened 2021-05-14
ParquetReader loses ticks


MaratFaskhiev:
**Version:** 3.8.6

Here is the test which reproduces behavior. Because NanoTime uses AddMilliseconds method it loses ticks during the conversion to DateTimeOffset type.

```csharp
      [Fact]
      public void ConvertToDateTimeOffset_PreservesTicks()
      {
         var dto = new DateTimeOffset(2021, 05, 14, 17, 52, 31, TimeSpan.Zero);
         dto = dto.Add(TimeSpan.FromTicks(1234567));
         var nanoTime = new NanoTime(dto);
         var convertedDto = (DateTimeOffset) nanoTime;

         Assert.Equal(dto, convertedDto);
      }
```

nilsgaeq:
This is still an issue for some date times, e.g. `2022-02-18T02:24:05.9104071+00:00"`. Happens because of rounding imprecision at ` _timeOfDayNanos = (long)(dt.TimeOfDay.TotalMilliseconds  * 1000000D);` in NanoTime constructor.

nilsgaeq:
Is there a reason one can't simply do `_timeOfDayNanos = dt.TimeOfDay.Ticks * 100` instead, as there are 100 ns in a tick?

aloneguid:
> Is there a reason one can't simply do `_timeOfDayNanos = dt.TimeOfDay.Ticks * 100` instead, as there are 100 ns in a tick?

I think this is a better way, you're correct. PR?

aloneguid:
closing as pr is merged

PR #112, opened 2021-05-14
Preserve ticks during the conversion from NanoTime to DateTimeOffset

MaratFaskhiev:
### Fixes

Issue #111 

### Description

Replace AddMilliseconds by AddTicks method to preserve ticks during the conversion from NanoTime to DateTimeOffset.


aloneguid (approved):
lgtm, thanks!

PR #113, opened 2021-05-16
Enable support for nullable Boolean

DBojsen:
### Fixes

Issue when Deserializing a file that has been saved with Python or C# including a nullable Boolean.
Before, it was mapped to wrong output (False -> True, True -> False) when Deserializing (in the generic IL code).

I am certain that the fix is not ready to be merged in current state, but wanted to raise PR to get awareness on issue,
and work with you to make the perfect fix.
The code in the current state maps correctly for both nullable and not nullable booleans:
Nullable Null -> False
Nullable True -> True
Nullable False -> False
False -> False
True -> True

### Description

* Inclusion of a NullableBooleanToBoolean converter.
* When detecting that data is in fact Nullable Boolean, change the CLR type on DataField (required making setter public)
* When invokink GenerateAssigner, use DataField definition saved on DataColumn instead of cashed version, which will be wrong for columns with Nullable Booleans

- [ No] I have included unit tests validating this fix.
- [ No] I have updated markdown documentation where required.
- [ Yes] I understand that successful approval of my pull request requires reproducible tests as per [Contribution Guideline](https://github.com/aloneguid/parquet-dotnet/blob/master/.github/CONTRIBUTING.md).

### DIFF
--- a/src/Parquet/Data/DataColumn.cs
+++ b/src/Parquet/Data/DataColumn.cs
@@ -68,7 +68,7 @@ namespace Parquet.Data
       /// <summary>
       /// Data field
       /// </summary>
-      public DataField Field { get; private set; }
+      public DataField Field { get; set; }
 
       /// <summary>
       /// When true, this field has repetitions. It doesn't mean that it's an array though. This property simply checks that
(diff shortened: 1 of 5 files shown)


aloneguid (approved):
lgtm

Issue #114, opened 2021-06-07
Int16 DataColumn in not of type Int16


BenjaminGopp:
CurrentVersion

#### Expected behavior

A DataColumn of DataType.Int16 should be of ClrType Int16, not null

#### Actual behavior

A DataColumn of DataType.Int16  ClrType is null

### Steps to reproduce the behavior

see test

#### Code snippet reproducing the behavior

```csharp
        [Test]
        public void Int16_Short()
        {
            var int16DataField = new DataField("Int16", DataType.Int16);
            var shortDataField = new DataField("Short", DataType.Short);

            Assert.That(int16DataField.DataType, Is.EqualTo(DataType.Int16));
            Assert.That(shortDataField.DataType, Is.EqualTo(DataType.Short));

            Assert.That(shortDataField.ClrType, Is.EqualTo(typeof(Int16)));
            Assert.That(int16DataField.ClrType, Is.EqualTo(typeof(Int16))); //null
        }
```

PR #116, opened 2021-06-12
 Fix for NullReference-Exception in case of Int16, UnsignedShort and UnsignedByte

nikolapeja6:
Copy of https://github.com/elastacloud/parquet-dotnet/pull/488

Adds/corrects the `DataType` to `BasicPrimitiveDataTypeHandler` mapping, solving the NullReference-Exception bug.

### Fixes

Issue #114

### Description

Not all values of the `DataType` enum had mappings to appropriate instances of `BasicPrimitiveDataTypeHandler` in `DataTypeFactory`, which I believe caused the unmatched types to be resolved to default objects in `_allDataTypes.FirstOrDefault(dt => dt.DataType == dataType)`, and that caused some of the fields of the `DataType` object to be null.

This PR adds and corrects `BasicPrimitiveDataTypeHandler`s so that every value of the `DataType` enum is mapped to an appropriate handler, solving the NullReference-Exception bug.

Fix is accompanied with a unit test which shows that objects created from all values of the `DataType` enum (except the `Unspecified` value) have a non-null `ClrType` field.

### DIFF
--- a/src/Parquet.Test/DataFieldTypeCoverageTests.cs
+++ b/src/Parquet.Test/DataFieldTypeCoverageTests.cs
@@ -1,0 +1,49 @@
+﻿using Parquet.Data;
+using System;
+using System.Collections.Generic;
+using System.IO;
+using System.Text;
+using Xunit;
+using Parquet.File.Values.Primitives;
+using NetBox.Extensions;
+using System.Linq;
+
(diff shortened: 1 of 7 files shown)


aloneguid (approved):
lgtm

PR #120, opened 2021-07-20
Fix: Issue with concurrent writing of string fields

rickykaare:
### Fixes

The 3.9.0 release introduced an issue with writing parquet files in parallel, where string field data would get mixed up with other string fields.

### Description

The issue has been reproduced by the `Flat_concurrent_write_read` unit test, and the cause has been identified as the latest changes to the `StringDataTypeHandler` where a static string buffer was introduced.

The suggested change is to revert the changes to `StringDataTypeHandler`.


ozeno (approved):

ThomasBoersma (approved):

JohniMichels:
This update saved me A LOT of work

PR #127, opened 2021-10-28
Add DbDataReader parquet writer

MarkPflug:
### Description

I started using this library yesterday for exporting data from SqlServer to parquet files. I ended up writing quite a bit of code to make it general purprose and thought it would be useful to contribute back to the project. This code is not SqlServer-specific, but targets the DbDataReader from the BCL which is standard data access class for all relational databases, including postgres, MySql, Sqlite, and myriad other libraries.

This allows writing code like the following (psedocode):

```C#
var conn = new SqlConnection(...);
var cmd = conn.CreateCommand();
cmd.CommandType = CommandType.StoredProcedure;
cmd.CommandText = "GetMyComplexDataSet";
DbDataReader reader = cmd.ExecuteReader();
reader.WriteParquet("MyComplexDataSet.parquet");

```

The changes are completely isolated from the rest of the project, so it would also be possible to move this to a separate project if needed.

The unit test uses my CSV parsing library which implements a strongly-typed DbDataReader, which allowed me to easily test using the existing csv files in the project. It would be possible to test without this library, but I would require adding a significant amount of code to the test library to implement the DbDataReader correctly.

This is a work-in-progress. I need to add more test coverage, specifically around row group size handling. This is intended to at least start a discussion about whether this would be an acceptable contribution.

I will likely also want to implement the inverse of this, which would be a DbDataReader implementation over parquet reader, which would allow doing things like easily bulk loading parquet into a relational database, or converting to CSV, etc.

I literally only started using parquet yesterday, so if I'm doing something obviously wrong I'd love to hear about it.

### DIFF
--- a/src/Parquet.Test/Extensions/DataReaderExtensionsTests.cs
+++ b/src/Parquet.Test/Extensions/DataReaderExtensionsTests.cs
@@ -1,0 +1,87 @@
+﻿using System;
+using System.Data.Common;
+using System.IO;
+using System.Linq;
+using Parquet.Data;
+using Parquet.Extensions;
+using Sylvan.Data.Csv;
+using Xunit;
+
+namespace Parquet.Test.Extensions
+{
+   public class DataReaderExtensionsTests
+   {
+      class CsvSchema : ICsvSchemaProvider
+      {
+         SchemaColumn[] columns;
+         public CsvSchema(params Type[] types)
+         {
+            this.columns = types.Select(t => new SchemaColumn(t)).ToArray();
+         }
+
+         public DbColumn GetColumn(string name, int ordinal)
+         {
+            return columns[ordinal];
+         }
+
+         class SchemaColumn : DbColumn
+         {
+            public SchemaColumn(Type type)
+            {
+               this.DataType = type;
+            }
+         }
+      }
+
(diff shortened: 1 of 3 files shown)


cajuncoding:
Looks like a Nice Extension that would be quite useful... hopefully it gets included!

MarkPflug:
The main issue with binding to objects then serializing is that it doesn't handle dynamic results sets, unless there's an aproach I'm not aware of. Using the DbDataReader directly avoids needing to know the shape of the data at compile time. It also allows creating parquet files for huge result sets that wouldn't fit in memory, since only a row groups worth of records would need to be resident in memory at a time.

cajuncoding:
In case it’s helpful, RepoDB does have decent support for Dynamics in both retrieval and storing into/from SQL databases:
https://repodb.net/feature/dynamics

aloneguid:
I think this is outside of this project's scope. It should be an external library or something like a parquet-dotnet recipe page in docs.

caseyoneill:
@MarkPflug Did you end up open sourcing this code as a standalone library?

MarkPflug:
@caseyoneill No, I did not. You can pull the code out of this PR though.

Issue #131, opened 2021-11-05
Support for Precision was a breaking (not-backwards) compatible change...


cajuncoding:
**Version:** Parquet.Net v3.8.1 vs 3.9

**Runtime Version:** .Net Core

**OS:** Windows/Linux

#### Expected behavior

The code compiles, and runs, but encounters runtime errors with c# decimal fields that have a `ParquetColumnAttribute` defined to map the column name!  It is expected that this would work the same as it always has, with default values for Precision, etc.!

#### Actual behavior
Unfortunately now runtime errors are encountered due to argument exceptions because 'precision' is not specified.

This is due to changes that allow for Precision to be defined on the attribute as metadata, but when not defined (which code being upgraded would not have this in place) the original defaults are no longer used, and argument validation errors are thrown at runtime. 

```
System.ArgumentException: 'precision cannot be less than 1 (Parameter 'precision')'
```

This is unlikely to be known by users of the library without release notes documenting the breaking change. 

This is expected to be a backwards compatible improvement whereby original default remain supported since there appears to be no reason to not allow original defaults to be used as they were up until v3.9!

In addition, if there is s driver to make these required then the original Attribute overload should be deprecated with clear details about which overload to use whereby the Attribute is actually required... or better design would have been to require the use of a different attirbute specifically for this purpose.

But, ideally it should simply be backwards compatible!

### Steps to reproduce the behavior

1. Have a model and a parquete file with Decimal values
2. Decorate the decimal property of the model with `ParquetColumnAttribute` _(without any of the newly supported Precision properties of the attribute)._
2. Deserialize your data via `var modelsWithDecimalProps = ParquetConvert.Deserialize<ModelWithDecimal>(stream);`

#### Code snippet reproducing the behavior
```csharp
using System;
using System.IO;
using System.Linq;
using System.Threading.Tasks;
using Parquet;
using Parquet.Attributes;


namespace ConsoleApp
{
    class Program
    {
        static async Task Main(string[] args)
        {
            await using var stream = new MemoryStream(await File.ReadAllBytesAsync("...path/to/sample/parquet/file"));
            var results = ParquetConvert.Deserialize<FCModel>(stream);

            var x = 1;
            Console.WriteLine($"[{results.Count()}] Valid Items found after Filtering!");
            foreach (var item in results)
            {
                //Console.WriteLine($"{x++}) {item.Id} -- {item.Category} [Budget={item.Budget}; InternalCost={item.InternalCost}]");
[...]

PR #132, opened 2021-11-06
Fix breaking change in v3.9 due to missing default fallback for Decimal Precision/Scale

cajuncoding:
### Fixes

Issue #131 

### Description

Fix breaking changes from v3.9 whereby new support for explicitly controlling Decimal precision and scale did not include support for previously existing default values. The changes introduced no longer allowed existing code to use defaults that were originally supported prior to v3.9.  

This is easily resolved by ensuring that the new properties available on `ParquetColumnAttribute` to provide clear default values when not explicitly defined by consumers.

The use case here is when we merely want to rename a `decimal` column like any other column; therefore this is now covered by the unit test `Serialise_deserialise_renamed_column` test.  The `Serialise_deserialise_renamed_column` test now passes successfully; but was failing during serialization & de-serialization when test data was added to the model.

Also factored out the default values for Precision & Scale into a common place in `DecimalFormatDefaults` class the and updated `DecimalDataTypeHandler` to also reference these defaults to eliminate duplication.

- [X] I have included unit tests validating this fix.
- [X ] I understand that successful approval of my pull request requires reproducible tests as per [Contribution Guideline](https://github.com/aloneguid/parquet-dotnet/blob/master/.github/CONTRIBUTING.md).

### DIFF
--- a/src/Parquet.Test/Serialisation/ParquetConvertTest.cs
+++ b/src/Parquet.Test/Serialisation/ParquetConvertTest.cs
@@ -149,7 +149,8 @@ namespace Parquet.Test.Serialisation
             .Select(i => new SimpleRenamed
             {
                Id = i,
-               PersonName = $"row {i}"
+               PersonName = $"row {i}",
+               NullableDecimal = (i % 3 == 0) ? null : (decimal)i / 3
             });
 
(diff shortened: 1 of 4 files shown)


aloneguid (approved):
lgtm

PR #150, opened 2022-02-07
Support LIsts for AWS Athena

bdquig:
I have a project that uses Parquet.Net to write files that are then used by AWS Athena.

Writing enumerables as an array DataField is not recognized by Athena.  Athena expects lists to be structured like:
  optional group col1 (LIST) {
    repeated group bag {
      optional binary array_element (STRING);
    }
  }
  optional group col2 (LIST) {
    repeated group bag {
      required int32 array_element;
    }
  }

I have updated ListField to allow setting the container name (rather than being hard coded to "list"). I also added a new constructor to allow setting the container name and the element name.  This constructor will simplify the setup by automatically creating the child DataField.  

`new ListField("col1", DataType.String, true, "stringEnumerable", "bag", "array_element");
new ListField("col2", DataType.Int32, false, "intEnumerable", "bag", "array_element");`

I also had to change the comparing of DataFields to use Path instead of Name since we can now have multiple DataFields with the same name.

TODO: I get a NullException if the enumerable is null when serializing.  As a workaround, I create the enumerable with a null element, and it does actually output a null.  I could not figure out how to update the generated IL to support this.  


aloneguid (approved):
lgtm

PR #177, opened 2022-09-21
FIX: Check if Year is 0, set it to 1 then

dvasseur:
NanoTime 1721424 equals to 0000-12-30 (YYYY-MM-DD) but DateTimeOffset.MinValue is 0001-01-01 => change Year to 1 (any better solution ?)


aloneguid (approved):
lgtm

Issue #179, opened 2022-09-27
Serialisation broken in 3.10.0 and 4.0.0 when a DataField's propertyName is null


El-Gor-do:
**Version:** Parquet.Net v3.10.0 and v4.0.0

**Runtime Version:** .Net 6.0

**OS:** Windows

#### Expected behavior

ParquetConvert.Serialize(...) should succeed if a schema's DataField is constructed with a null propertyName.

#### Actual behavior

ParquetConvert.Serialize(...) throws ArgumentNullException.

#### Code snippet reproducing the behavior

Sample console app using Parquet 3.9.1 that successfully serialises.
```
using System.Collections.Generic;
using System.IO;

using Parquet;
using Parquet.Data;

namespace TestParquet
{
    internal class Program
    {
        public class TestClass
        {
            public int Int32Value { get; set; }

            public static Schema CreateSchema()
            {
                return new Schema(
                    new DataField(nameof(Int32Value), DataType.Int32, false, false)
                    );
            }
        }

        static void Main()
        {
            List<TestClass> items = new List<TestClass>()
            {
                new TestClass() { Int32Value = 1 },
                new TestClass() { Int32Value = 2 },
            };

            string filePath = Path.GetTempFileName();
            const int bufferSize = 16384;
            using (FileStream stream = new FileStream(filePath, FileMode.Create, FileAccess.ReadWrite, FileShare.Read, bufferSize, FileOptions.DeleteOnClose))
            {
                Schema schema = TestClass.CreateSchema();
                CompressionMethod compressionMethod = CompressionMethod.Gzip;
                const int rowGroupSize = 5000;

                schema = ParquetConvert.Serialize(items, stream, schema, compressionMethod, rowGroupSize);
            }
        }
    }
}
```

Update the Parquet.Net reference to 3.10.0 and run again. The last line `schema = ParquetConvert.Serialize(items, stream, schema, compressionMethod, rowGroupSize);` now throws
```
Unhandled exception. System.ArgumentNullException: Value cannot be null. (Parameter 'name')
   at System.Type.GetProperty(String name, BindingFlags bindingAttr)
   at System.Reflection.TypeInfo.GetDeclaredProperty(String name)
   at Parquet.Serialization.Values.MSILGenerator.GenerateCollector(ILGenerator il, Type classType, DataField f, MethodInfo getValueMethod, MethodInfo addToListMethod, MethodInfo addRepLevelMethod, TypeConversion conversion)
[...]

PR #180, opened 2022-09-28
Fix ArgumentNullException during serialisation when schema data field…

El-Gor-do:
… is constructed with null propertyName

### Fixes

Issue #179 

### Description

If a schema is manually constructed using DataFields where propertyName == null then ParquetConvert.SerializeAsync(...) would throw an ArgumentNullException. This PR fixes that bug and also adds propertyName to the DateTimeDataField and DecimalDataField constructor arguments.


aloneguid (approved):
lgtm

PR #186, opened 2022-10-12
ReadAsTableAsync: Return an empty table if there is no row

dvasseur:
Avoid an error when using `foreach (Row row in parquetReader.ReadAsTable())`for example


aloneguid (approved):
lgtm

PR #189, opened 2022-10-20
Fixed #188 - Reading full page from stream

### LINKED ISSUE #188: Bug in DataColumnReader.ReadPageData when using Azure Data Lake SDK
**Version:** Parquet.Net v4.1.0

**Runtime Version:** .Net 6

**OS:** Windows

#### Expected behavior

ReadPageData should read whole page

#### Actual behavior

It only reads the first bytes returned from _inputStream.ReadAsync

#### Why this happens

The [Stream.Read](https://learn.microsoft.com/en-us/dotnet/api/system.io.stream.read?view=net-6.0) interface specifies that Read can return less than the supplied count:
> The total number of bytes read into the buffer. This can be less than the number of bytes allocated in the buffer if that many bytes are not currently available, or zero (0) if the end of the stream has been reached.

Azure Data Lake SDK [LazyLoadingReadOnlyStream.ReadInternal](https://github.com/Azure/azure-sdk-for-net/blob/main/sdk/storage/Azure.Storage.Common/src/Shared/LazyLoadingReadOnlyStream.cs#L224-L234
) does this when the requested number of bytes is larger than the number of bytes remaining in the buffer.

[DataColumnReader.ReadPageData](https://github.com/aloneguid/parquet-dotnet/blob/67173d7534240f65fb4ced0372d8ef6bc4459a2b/src/Parquet/File/DataColumnReader.cs#L140) only calls `_inputStream.ReadAsync` once and does not check its return value. It should keep calling it as long as it returns > 0 and not all requested data has been returned yet. Similar to what is done [here in Azure Data Lake SDK](https://github.com/Azure/azure-sdk-for-net/blob/main/sdk/storage/Azure.Storage.Common/src/Shared/LazyLoadingReadOnlyStream.cs#L265-L286)

### Steps to reproduce the behavior

1. Create a DataLakeFileClient
2. Open a read stream with a small buffer size (e.g. 256kb)
3. Read a parquet file where the compressed page size > 256kb

#### Code snippet reproducing the behavior

The issue can also be reproduced by using a dumb stream implementation that only returns one byte at the time:

```csharp
using Parquet;
using Parquet.Data;

internal class Program
{
    private static async Task Main(string[] args)
    {
        DataField field = new("date", typeof(DateTimeOffset));
        Schema schema = new(field);

        using MemoryStream memoryFileStream = new();
        using (ParquetWriter parquetWriter = await ParquetWriter.CreateAsync(schema, memoryFileStream))
        {
            using ParquetRowGroupWriter rowGroupWriter = parquetWriter.CreateRowGroup();
            DateTimeOffset[] data = new DateTimeOffset[10000];
            for (int i = 0; i < 10000; i++) data[i] = DateTimeOffset.UtcNow.AddMilliseconds(i);
            await rowGroupWriter.WriteColumnAsync(new DataColumn(field, data, 0, 10000));
        }

        int fileSize = (int)memoryFileStream.Length;
        using MemoryStream memoryStreamCopy = new(memoryFileStream.GetBuffer(), 0, fileSize);
        using DumbStream bufferedStream = new(memoryStreamCopy);
        using ParquetReader parquetReader = await ParquetReader.CreateAsync(bufferedStream);
        for (int iterations = 0; iterations < 100; iterations++)
        {
            using ParquetRowGroupReader rowGroupReader = parquetReader.OpenRowGroupReader(0);
            await rowGroupReader.ReadColumnAsync(field);
        }
    }

    internal class DumbStream : Stream
    {
        private readonly MemoryStream _memoryStream;
        public DumbStream(MemoryStream memoryStream) { _memoryStream = memoryStream; }
        public override bool CanRead => true;
        public override bool CanSeek => true;
        public override bool CanWrite => false;
        public override long Length => _memoryStream.Length;
        public override long Position { get => _memoryStream.Position; set => _memoryStream.Position = value; }
        public override void Flush() { }
        public override int Read(byte[] buffer, int offset, int count) => _memoryStream.Read(buffer, offset, 1);
        public override long Seek(long offset, SeekOrigin origin) => _memoryStream.Seek(offset, origin);
        public override void SetLength(long value) => throw new NotSupportedException();
        public override void Write(byte[] buffer, int offset, int count) => throw new NotSupportedException();
        public override Task<int> ReadAsync(byte[] buffer, int offset, int count, CancellationToken cancellationToken) => Task.FromResult(Read(buffer, offset, count));
    }
}
```

Sense545:
### Fixes

Issue #188

### Description

- Fixed bug in ReadPageData where it did not read all the page bytes from the stream

### DIFF
--- a/src/Parquet.Test/DumbStreamTest.cs
+++ b/src/Parquet.Test/DumbStreamTest.cs
@@ -1,0 +1,51 @@
+using System;
+using System.IO;
+using System.Threading;
+using System.Threading.Tasks;
+using Parquet.Data;
+using Xunit;
+
+namespace Parquet.Test {
+    public class DumbStreamTest {
+        [Fact]
+        public async Task Read_from_stream_that_only_retuns_one_byte_at_the_time() {
+            DataField field = new("date", typeof(DateTimeOffset));
+            Schema schema = new(field);
+
+            using MemoryStream memoryFileStream = new();
+            using(ParquetWriter parquetWriter = await ParquetWriter.CreateAsync(schema, memoryFileStream)) {
+                using ParquetRowGroupWriter rowGroupWriter = parquetWriter.CreateRowGroup();
+                DateTimeOffset[] data = new DateTimeOffset[10000];
+                for(int i = 0; i < 10000; i++)
+                    data[i] = DateTimeOffset.UtcNow.AddMilliseconds(i);
+                await rowGroupWriter.WriteColumnAsync(new DataColumn(field, data, 0, 10000));
+            }
+
+            int fileSize = (int)memoryFileStream.Length;
+            using MemoryStream memoryStreamCopy = new(memoryFileStream.GetBuffer(), 0, fileSize);
+            using DumbStream bufferedStream = new(memoryStreamCopy);
+            using ParquetReader parquetReader = await ParquetReader.CreateAsync(bufferedStream);
+            for(int iterations = 0; iterations < 100; iterations++) {
+                using ParquetRowGroupReader rowGroupReader = parquetReader.OpenRowGroupReader(0);
+                await rowGroupReader.ReadColumnAsync(field);
+            }
+        }
+
+        private class DumbStream : Stream {
+            private readonly MemoryStream _memoryStream;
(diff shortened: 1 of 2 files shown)


aloneguid (approved):
lgtm

Issue #201, opened 2022-12-01
Serialization Currently Misses Base Class / Inherited Properties


DrewMcArthur:
**Version:** Parquet.Net v4.1.0

**Runtime Version:** .NET6.0

**OS:** Windows

#### Expected behavior

We have a base class with property `A`, and an inherited class with property `B`.  When serializing an instance of the inherited class, we expected the Parquet output to have both properties `A` and `B`.

#### Actual behavior

It appears that the serialize function ignores properties in the base class, so the parquet output only has property `B`.  [We've noticed that this is a noted limitation here](https://github.com/aloneguid/parquet-dotnet/blob/master/docs/serialisation.md#limitations), but this would be nice to have.  Would be happy to work towards this if you could point us in the right direction!

### Steps to reproduce the behavior

1. Create a base class for serialization.
```csharp
public class BaseClass 
{
    [ParquetColumn(Name="property_a")]
    public string PropertyA;
}
```
2. Create another class for serialization, inheriting from the base class.
```csharp
public class InheritedClass : BaseClass
{
    [ParquetColumn(Name="property_b")]
    public string PropertyB;
}
```
3. Create an instance of the inherited class and serialize it.
```csharp
InheritedClass obj = new() 
{
    PropertyA = "a";
    PropertyB = "b";
}
MemoryStream memStream = new();
await ParquetConvert.SerializeAsync(obj, memStream, ...);
```
4.  Check the parquet file, and only see `PropertyB` in the output.  

#### Code snippet reproducing the behavior

I think what I've provided so far is enough to go off of - would be happy to provide more context / a unit test in a PR if you'd like!  Since I found this limitation is mentioned in your documentation, but didn't see any github issues mentioning it, I figured I'd open the issue to track progress.  Thanks for all your work on this library!

DrewMcArthur:
going to start this by opening up a PR with a unit test describing the issue - will link back here once i do 😄

DrewMcArthur:
closed with the merge of #202 ! thanks @aloneguid !

PR #202, opened 2022-12-02
allow serialization of inherited properties

DrewMcArthur:
### Fixes

Issue #201 

### Description

adds methods for schema creation that includes properties inherited from the given type's base class


src/Parquet/Serialization/Values/MSILGenerator.cs
```diff
@@ -385,4 +385,34 @@ public static Boolean NullableBooleanToBooleanConversion(Boolean? value)
...
+   static class PropertyHelpers {
+      /// <summary>
+      /// when the field is declared on the base class, the original version produced 
+      /// a NullReferenceException.  this helper method will fetch the property off the base class
+      /// if necessary, preventing the NRE
+      /// </summary>
+      /// <param name="classType">Type</param>
+      /// <param name="field">Field</param>
+      /// <returns>PropertyInfo</returns>
+      public static PropertyInfo GetDeclaredPropertyFromClassType(Type classType, Field field)
```
DrewMcArthur, line 402: open to alternative name suggestions for this function; also if it should be put somewhere else!

src/Parquet/Serialization/SchemaReflector.cs
```diff
@@ -98,5 +151,7 @@ private Field GetField(PropertyInfo property)
 
       Func<PropertyInfo, bool> pickSerializableProperties = (PropertyInfo arg) => !arg.CustomAttributes.Any(p => p.AttributeType == typeof(ParquetIgnoreAttribute));
 
+      // TODO: initially defined this using the pattern above, but didn't know the syntax for a type-generic version, so wrote this instead.  any issue with that?
+      private static bool isNotNull<T>(T obj) => obj != null;
```
DrewMcArthur, line 155: see todo - is this pattern an okay way to do this, or is there a generically-typed way to have a ’Func’ type variable, like the line above this?

src/Parquet/Serialization/SchemaReflector.cs
```diff
@@ -59,6 +87,31 @@ public static Schema Reflect(Type classType)
...
+
+      /// <summary>
+      /// see non-static method <see cref="ReflectWithInheritedProperties()"/>
+      /// </summary>
+      /// <param name="classType"></param>
+      /// <returns><see cref="Schema"/></returns>
+      public static Schema ReflectWithInheritedProperties(Type classType)
+      {
+         // TODO: how to differentiate between cached schemas for a type with/without inherited properties.  different cache?
+         return _cachedReflectedSchemas.GetOrAdd(
```
DrewMcArthur, line 108: i'm thinking there should just be a separate cache.  a little clutter in the code having an extra instance variable, but shouldn't hurt performance at all.

another option would be to make the cache dictionary key a tuple with some modifier, but that feels more complex.

src/Parquet/Serialization/SchemaReflector.cs
```diff
@@ -59,6 +87,31 @@ public static Schema Reflect(Type classType)
          return _cachedReflectedSchemas.GetOrAdd(classType, t => new SchemaReflector(classType).Reflect());
       }
 
+      /// <summary>
+      /// see non-static method <see cref="SchemaReflector.ReflectWithInheritedProperties()"/>
```
DrewMcArthur, line 91: i'm not sure if i did this cref correctly, so i'll revisit that to make sure.

src/Parquet/Serialization/SchemaReflector.cs
```diff
@@ -39,6 +39,34 @@ public Schema Reflect()
...
+      ///         which, if not null, is used to add to the property list in the normal method. 
+      ///         then, the static methods would just 
+      ///            `new SchemaReflector(classType, baseClassType).Reflect()`
+      ///         instead of what we have now.
+      /// </summary>
+      /// <returns><see cref="Schema"/></returns>
+      public Schema ReflectWithInheritedProperties()
+      {
+         IEnumerable<PropertyInfo> properties = _classType.DeclaredProperties;
+         // TODO: BaseClass isn't a valid property here
```
DrewMcArthur, line 57:
```suggestion
```

src/Parquet/ParquetConvert.cs
```diff
@@ -97,13 +97,18 @@ public static class ParquetConvert {
...
         /// <returns></returns>
-        public static async Task<T[]> DeserializeAsync<T>(Stream input, int rowGroupIndex = -1) where T : new() {
+        public static async Task<T[]> DeserializeAsync<T>(Stream input, int rowGroupIndex = -1, 
+            Schema fileSchema = null
+        ) where T : new() {
             var result = new List<T>();
             using(ParquetReader reader = await ParquetReader.CreateAsync(input)) {
-                Schema fileSchema = new SchemaReflector(typeof(T)).Reflect();
+                if (fileSchema == null) {
+                    fileSchema = new SchemaReflector(typeof(T)).Reflect();
```
DrewMcArthur, line 110:
```suggestion
                    fileSchema = fileSchema ?? SchemaReflector.Reflect<T>();
```
pseudo code, but do this instead of if statement?

src/Parquet.Test/InheritedPropertiesTest.cs
```diff
@@ -0,0 +1,55 @@
...
+      MemoryStream stream = new();
+      await ParquetConvert.SerializeAsync(recordsToSerialize, stream, schema);
+      InheritedClass[] deserializedRecords = await ParquetConvert.DeserializeAsync<InheritedClass>(stream, fileSchema: schema);
+      
+      InheritedClass expected = recordsToSerialize[0];
+      InheritedClass actual = deserializedRecords[0];
+
+      // TODO: should we instead assert that 
+      //    `deserialized[0].BaseProperty is not null`?
+      Assert.Equal(expected.BaseProperty, actual.BaseProperty);
```
DrewMcArthur, line 44: how do we feel about this check - should we be more thorough?

fyi, existing behavior before this PR is that ’expected.BaseProperty == null’

DrewMcArthur (commented):
some comments for me to get back to, as well as some for input from a maintainer/someone with more c# experience.

src/Parquet/Serialization/SchemaReflector.cs
DrewMcArthur: added unit tests for this; going to go with the first option (a separate cache).

because c# lets you inherit multiple layers; i'm now wondering if a better implementation would be an overload like this, maybe a `Type[]` to allow multiple base classes.

```csharp
public static Schema Reflect(Type classType, Type baseClassType) ...
```

src/Parquet.Test/InheritedPropertiesTest.cs
DrewMcArthur: these tests were updated so this comment is outdated i'm pretty sure.

src/Parquet.Test/InheritedPropertiesTest.cs
```diff
@@ -0,0 +1,66 @@
...
+
+    [Fact]
+    public async Task Serialize_class_without_inherited_properties() {
+      InheritedClass[] recordsToSerialize = GenerateRecordsToSerialize();
+      Schema schema = SchemaReflector.Reflect<InheritedClass>();
+
+      // TODO: is there some `with/using` syntax i should use here?
+      MemoryStream stream = new();
+      await ParquetConvert.SerializeAsync(recordsToSerialize, stream, schema);
+      InheritedClass[] deserializedRecords = await ParquetConvert.DeserializeAsync<InheritedClass>(stream, fileSchema: schema);
```
DrewMcArthur, line 31:
```suggestion
      MemoryStream stream = new();
      await ParquetConvert.SerializeAsync(recordsToSerialize, stream, schema);
      InheritedClass[] deserializedRecords = await ParquetConvert.DeserializeAsync<InheritedClass>(stream, fileSchema: schema);
```
opening up a convo around this syntax - i'm new to c#, so wasn't sure if i should be utilizing some `using` keyword with these variables.  let me know what the best practice is!

src/Parquet/Serialization/SchemaReflector.cs
```diff
@@ -39,6 +40,33 @@ public Schema Reflect()
          return new Schema(properties.Select(GetField).Where(p => p != null).ToList());
       }
 
+      /// <summary>
+      /// TODO: i think this is the wrong cref.  how to point to non-static instance method?
+      /// Same functionality as <see cref="SchemaReflector.Reflect()"/>,
```
DrewMcArthur, line 45: todo: update cref with correct syntax

src/Parquet/Serialization/SchemaReflector.cs
```diff
@@ -39,6 +40,33 @@ public Schema Reflect()
...
+      /// <summary>
+      /// TODO: i think this is the wrong cref.  how to point to non-static instance method?
+      /// Same functionality as <see cref="SchemaReflector.Reflect()"/>, 
+      /// but this method includes any `DeclaredProperties` inherited from 
+      /// the given `classType`'s `BaseClass`
+      /// TODO: instead of doing things this way, we can add a nullable "_baseClassType" attribute,
+      ///         which, if not null, is used to add to the property list in the normal method. 
+      ///         then, the static methods would just 
+      ///            `new SchemaReflector(classType, baseClassType).Reflect()`
+      ///         instead of what we have now.
```
DrewMcArthur, line 52: this is the other option i mentioned in the conversation around caching - thoughts?

src/Parquet/Serialization/SchemaReflector.cs
```diff
@@ -39,6 +40,33 @@ public Schema Reflect()
...
+      /// </summary>
+      /// <returns><see cref="Schema"/></returns>
+      public Schema ReflectWithInheritedProperties()
+      {
+         IEnumerable<PropertyInfo> properties = _classType.DeclaredProperties;
+         IEnumerable<PropertyInfo> baseClassProperties = _classType.BaseType.GetTypeInfo().DeclaredProperties;
+         // TODO: can we just chain the addrange into the LINQ below?
+         IEnumerable<PropertyInfo> allProperties = baseClassProperties.Concat(properties);
+
+         List<Field> allValidFields = allProperties.Where(pickSerializableProperties)
```
DrewMcArthur, line 62:
```suggestion


         List<Field> allValidFields = baseClassProperties.Concat(properties)
                                                                                          .Where(pickSerializableProperties)
```

src/Parquet/Serialization/SchemaReflector.cs
```diff
@@ -15,6 +15,7 @@ public class SchemaReflector
    {
       private readonly TypeInfo _classType;
       private static readonly ConcurrentDictionary<Type, Schema> _cachedReflectedSchemas = new ConcurrentDictionary<Type, Schema>();
+      private static readonly ConcurrentDictionary<Type, Schema> _cachedReflectedWithInheritedPropertiesSchemas = new();
```
DrewMcArthur, line 18: it seems that the public constructors and the static methods are used interchangeably around the codebase - however the constructors skip these caches (since the caching happens in the static methods).

is that intentional, or should the constructors be made private, to encourage using the cache?

src/Parquet/ParquetConvert.cs
DrewMcArthur: also, should we be using the static methods over the public constructor, or is there a reason we don't want to cache this schema?  see my comment [here](https://github.com/aloneguid/parquet-dotnet/pull/202#discussion_r1038295559)

src/Parquet/Serialization/SchemaReflector.cs
DrewMcArthur: updated, but still not sure.  basically we have a method `Reflect()`, a `static Reflect()` and `static Reflect(Type)`.  ideally i'd like to reference the non-static method, but i couldn't find how to do that.

src/Parquet/Serialization/SchemaReflector.cs
DrewMcArthur: resolved this by adding another cache variable, but leaving this open so that a maintainer sees this convo.

MarkEnverus:
Has there been any movement on this?  This would be very helpful

DrewMcArthur:
@MarkEnverus hi mark, glad this would be useful for you! no movement; i haven't heard from @aloneguid on this.  not sure if they saw it, or what the right method for getting their attention is.  my use case has a workaround for now, so pushing this through hasn't been my priority for now.

MarkEnverus:
What’s your workaround?  I’m happy to use it to move forward
________________________________

DrewMcArthur:
we had a very simple base class that was being serialized, so we just copied the inherited properties into the child classes.  ends up being some repetitive code, but not the worst solution.

MarkEnverus:
we did the same!  just duplicative code inside private classes - not elegant but effective

aloneguid (approved):
thanks, looking good! Pushing into v4.3.1

DrewMcArthur:
@aloneguid sweet, thank you!! glad it looked good to you, and thanks for updating for your latest version.  
@MarkEnverus making sure you saw this just got merged, you can target these changes with v4.2.1 in your csproj file!

MarkEnverus:
Thanks!!

MarkEnverus:
I tested and i'm till seeing issues - did this work for @DrewMcArthur?

DrewMcArthur:
> I tested and i'm till seeing issues - did this work for @DrewMcArthur?

@MarkEnverus i haven't gotten the chance to implement in our project that'd use this yet - have you seen how the new functionality is used in the new unit tests? i'd use that as a guide for how to get inherited properties serialized right - once you've done that, i'd be happy to look closer if you put more info in these comments or email me directly, up to you.

MarkEnverus:
my bad @DrewMcArthur  i wasn't passing the schema (i didn't use the schema previously) just assumed it worked... apologies... seems to work w/ the schema!

DrewMcArthur:
awesome @MarkEnverus !! glad to hear it.

sorry for the confusion - i did add some documentation in the repo about inherited properties, so hopefully that's helpful for people trying to figure this out in the future.

do you mind taking a look at those docs and checking what i wrote, to see if it makes sense? there should be an `inheritance.md` file somewhere...

MarkEnverus:
Docs look great @DrewMcArthur  - my fault for not reading them.

Issue #203, opened 2022-12-02
Support .netstandard 2.0


achen831008:
**Version:** Parquet.Net v4.1.2

**Runtime Version:** .Net Framework v4.8

**OS:** Windows

#### Expected behavior

Install Parquet.Net to .net framework 4.8 based project

#### Actual behavior

latest Parquet.Net no longer support .netstandard 2.0

aloneguid:
It will be a lot of work to support old standard.

PR #204, opened 2022-12-02
Support .netstandard 2.0

achen831008:
I need to the latest version to support .net standard 2.0

### Fixes

Issue #203 

### Description

desctiption goes here

- [ ] I understand that successful approval of my pull request requires reproducible tests as per [Contribution Guideline](https://github.com/aloneguid/parquet-dotnet/blob/master/.github/CONTRIBUTING.md).


aloneguid:
It's not just a matter of changing a number, you'll have to implement missing runtime parts so it won't work

Issue #235, opened 2023-01-24
Add option to decide compression level of for instance gzip compression


dxdjgl:
Changing from parquet.net version 3.10 to 4.3.4 has introduced increased cpu usage due to the fact that gzip is now using smallestsize instead of optimal compression. It would be very to have the freedom to select according to own usecase.

dxdjgl:
I have made a pull request on ironcompress to support this feature, I will make pull request when IronCompress has been accepted and new nuget package is available.

Issue #221, opened 2023-01-12
🈁💌💗 Post your use cases / software / screenshots here


aloneguid:
Hi everyone! This is an open thread for posting examples of the way you are using `Parquet.Net`. I'm looking for real world usage!

For example:

> I use `Parquet.Net` to store and query my collection of memes. It's very fast and efficient, and it supports all kinds of data types, like strings, integers, Booleans, and even emojis! 😂

Of course it's a joke, but you get the idea! 

How do you use `Parquet.Net`? Share your examples below, and don't forget to check you are not violating your employer's privacy policy, intellectual rights and so on. You don't want to end up like Bob, who accidentally leaked the company's secret formula for their best-selling product. He thought he was just sharing a harmless photo of his lunch, but he didn't notice the paper on his desk with the confidential information. Now he is looking for a new job and a new place to eat.

aloneguid:
[ML.NET](https://github.com/dotnet/machinelearning) seems to be using Parquet.Net in order to decode files, reference to [.csproj](https://github.com/dotnet/machinelearning/blob/main/src/Microsoft.ML.Parquet/Microsoft.ML.Parquet.csproj).

aloneguid:
[ParquetViewer](https://github.com/mukunku/ParquetViewer) - native Windows Forms app.

![image](https://user-images.githubusercontent.com/3155189/212433149-4e50a373-ca40-41a1-8e4a-d47ebd264468.png)

mithom:
we use it to store data gathered in testcell tests. It forms the backend of our data sharing/storing and processing framework around testing. When implemented correctly (still WIP), it would allow our calculation engineers to search through all the test results and filter based on some of the data of the test results

aloneguid:
<img width="872" alt="image" src="https://user-images.githubusercontent.com/3155189/212929230-ca1a48ba-3e36-4a3b-89e7-23e96a300120.png">

[Recfluence](https://recfluence.net/) seems to be using parquet.net to [read data](https://github.com/markledwich2/Recfluence/blob/f1b1ff83cdfcab85290aa05fa9a5eeacf023960c/App/YtReader/YtReader.csproj).

aloneguid:
[...]

PR #242, opened 2023-01-26
Add option to decide compression level of for instance gzip compressi…

dxdjgl:
…on #235

### DIFF
--- a/src/Parquet/File/Compressor.cs
+++ b/src/Parquet/File/Compressor.cs
@@ -1,11 +1,12 @@
 ﻿using System;
+using System.IO.Compression;
 using IronCompress;
 
 namespace Parquet.File {
     static class Compressor {
         private static readonly Iron _iron = new();
 
-        public static DataBuffer Compress(CompressionMethod method, ReadOnlySpan<byte> input) => _iron.Compress(ToCodec(method), input);
+        public static DataBuffer Compress(CompressionMethod method, ReadOnlySpan<byte> input, CompressionLevel compressionLevel) => _iron.Compress(ToCodec(method), input, compressionLevel: compressionLevel);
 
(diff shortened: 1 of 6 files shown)


aloneguid (changes requested):
Looks good to me. Worth updating to upcoming 1.3 of IronCompress though.

dxdjgl:
I will change to that version right away, since I was also complaining about the performance drop due to managed snappy.

aloneguid:
> I will change to that version right away, since I was also complaining about the performance drop due to managed snappy.

Thanks. The author of managed snappy was promising near-native performance, but i haven't benchmarked it ;)

dxdjgl:
With my data it is 5 times slower, from ~100 second archiving to > 500 second, so for me a big step backwards.

aloneguid:
would be interested to hear your use case in #221 if not a secret ;)

aloneguid:
@dxdjgl IronCompress 1.3 is out, will take a few minutes for nuget.org to index it.

aloneguid:
btw you might want to try Snappy compression instead (the default). It's faster and recent addition of logical dictionary compression will make output size similar.

dxdjgl:
> btw you might want to try Snappy compression instead (the default). It's faster and recent addition of logical dictionary compression will make output size similar.

Similar to GZip, and better performance that sounds very interesting. I dont have much cpu but quite a bit of disk.

dxdjgl:
@aloneguid I would rather not do the update to IronCompress 1.3 since there are some breaking changes. If you think this is fine please merge it and you could have a look at the IronCompress update.

aloneguid:
That's fine, breaking changes are not surfacing out of parquet lib.

dxdjgl:
> That's fine, breaking changes are not surfacing out of parquet lib.

I dont feel confident in integrating the changes so I would ask you to perform the updates.

dxdjgl:
Thanks for spending time integrating the new IronCompress version, it will give me back performance I need.

aloneguid (approved):

PR #248, opened 2023-01-30
Reading back the schema with adjusted encoding of DateTime, is not re…

dxdjgl:
…flected correctly.

After saving a schema which contains DateTime where only DateAndTime is stored not "Impala", reading back the schema from the file will indicate Impala.

I just want to know if this is by design or an error, I have made changes in my code to workaround this issue. But perhaps this should be solved in the Parquet.net library instead.


aloneguid:
Yeah you are absolutely correct on this.

aloneguid (approved):
great fix, merging.

aloneguid (approved):

PR #250, opened 2023-01-31
Changes to test code to run on my machine

xclayl:
I'm not sure if stats DateTime behaviour is what you want (for non-GMT TZs).
The other change allows for spaces in path names.


aloneguid (approved):
Lgtm

PR #253, opened 2023-02-07
Testing the 'Unspecified Kind' behaviour (and Local behaviour)

xclayl:
I don't expect this to be merged yet, but probably start a discussion.  Whatever is decided, it's probably good to test the wanted behaviour.

Looks like the storage of DateTime values are in UTC.  (hopefully I got this right)

What should happen when the code is given a DateTime that is either Local or Unspecified?
Local dates work fine 99% of the time, except once a year when an hour repeats.  So the code can convert it to UTC most of the time.
Unspecified dates are a bit more complicated.  Do you assume it is UTC or Local?  The code in this PR assumes the 'unspecified' DateTime values are in UTC. 
This code bets a developer will care about Local vs UTC that is stored in a parquet file.  05/06/2020 03:59 UTC = 04/06/2020 22:59 EST (ie different days), so the date given probably should match the date stored.


aloneguid:
Yeah it assumes everything is in UTC as most of the time this code runs on a server and "local" does not make much sense. I guess you are right, we can merge it.

aloneguid (dismissed):

aloneguid (approved):

PR #259, opened 2023-02-10
Fix issue destroying existing file with append in ParquetConvert.Seri…

oyhovd:
Fix issue destroying existing file with append in `ParquetConvert.SerializeAsync()`

Previous version would overwrite any existing file, even when the append flag was set to true due to the way the file was opened with `File.Create()`. Changed to use `File.Open()`.

### DIFF
--- a/src/Parquet.Test/Serialisation/ParquetConvertTest.cs
+++ b/src/Parquet.Test/Serialisation/ParquetConvertTest.cs
@@ -141,6 +141,141 @@ namespace Parquet.Test.Serialisation {
             }
         }
 
+        /// <summary>
+        /// Write to a new file, then append to it. Expect to deserialize it successfully.
+        /// </summary>
+        [Fact]
+        public async Task Serialize_new_file_append_deserialize() {
+            DateTime now = DateTime.Now;
+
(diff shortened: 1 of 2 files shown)


aloneguid (changes requested):
Might want to add tests for this, otherwise it will be broken again at some point in the future.

oyhovd:
@aloneguid Added some tests.

A couple of questions that came up while writing tests:

1. What is the intended behavior when calling `SerializeAsync()` with `append: true` when there are no pre-existing file? Currently there is some cryptic IOException: "only 0 out of 4 bytes are available". Should it work, i.e. as appending to a file with 0 existing rows, or throw a more specific exception?

2. What is the intended behavior when when calling  `SerializeAsync()` with `append: false` when the file exists, i.e. when wanting to overwriting the file, and the schema (object type) is different from that in the existing file? Should the schema still be expected to match and an exception be thrown, or the existing file just be wiped? Currently the writing goes through, but an undescriptive `Thrift.Protocol.TProtocolException` is thrown with no message when trying to deserialize the file with `DeserializeAsync()` after, regardless of whether I try to deserialize to the former or latter object type.

aloneguid (approved):
@oyhovd

Issue #282, opened 2023-03-26
[BUG]: Nested nulls can be interpreted at the wrong level 


jmandel:
### Library Version

4.6.1

### OS

Linux

### OS Architecture

64 bit

### How to reproduce?

I'm encountering problems with serialization of nested null values. Results of serialization->deserialiation can interpret the null at the wrong level.

I encountered issues working with the Table API  with nested arrays, but when I tried to put together a minimal breaking example, I found this relatively simple case that leads to failure with the Class API, without arrays.

### Failing test

Try this test in `ParquetSerializerTest.cs`:

```csharp
        [Fact]
        public async Task Struct_Serde_With_NestedNulls() {

            var data = new List<AddressBookEntry> {
                new AddressBookEntry {
                    FirstName = "Joe",
                    LastName = "Bloggs",
                    Address = new Address() {
                        City = null,
                        Country = null
                    }
                }
            };

            using var ms = new MemoryStream();
            await ParquetSerializer.SerializeAsync(data, ms);

            ms.Position = 0;
            IList<AddressBookEntry> data2 = await ParquetSerializer.DeserializeAsync<AddressBookEntry>(ms);

            Assert.Equivalent(data2, data);
        }
```

This tests should succeed, but it fails with

```sh
[xUnit.net 00:00:01.16]     Parquet.Test.Serialisation.ParquetSerializerTest.Struct_Serde_With_NestedNulls [FAIL]
  Failed Parquet.Test.Serialisation.ParquetSerializerTest.Struct_Serde_With_NestedNulls [269 ms]
  Error Message:
   Assert.Equivalent() Failure: Collection value not found
Expected: AddressBookEntry { Address = null, FirstName = "Joe", LastName = "Bloggs" }
In:       [AddressBookEntry { Address = Address { ... }, FirstName = "Joe", LastName = "Bloggs" }]
  Stack Trace:
     at Parquet.Test.Serialisation.ParquetSerializerTest.Struct_Serde_With_NestedNulls() in /tmp/parquet-dotnet/src/Parquet.Test/Serialisation/ParquetSerializerTest.cs:line 118
--- End of stack trace from previous location ---

Failed!  - Failed:     1, Passed:     0, Skipped:     0, Total:     1, Duration: < 1 ms - Parquet.Test.dll (net7.0)
```

thomas-bios:
@aloneguid I've updated my comment, added the repetition level of the column and a precision.

[...]

PR #285, opened 2023-03-27
Add tests about repeated field with nulls

thomas-bios:
See issue #282


thomas-bios:
@aloneguid (With surprises) these two tests are not failing, but I think I understand the issue that I have with my use case. See the comment in the new test `Repeated_field_with_nulls_write_read`

 I'll continue digging to see from where it is coming.

aloneguid:
closing as this was resolved

PR #333, opened 2023-05-18
Added missing flush, which makes my code hang, the flush was present …

dxdjgl:
…in version 4.9.0 but removed in 4.9.1.


aloneguid (approved):
looks good, thanks!

PR #351, opened 2023-06-15
Added backward compatibility for parquets with array primitives collection serialzied via legacy ParquetConverter

PablitoCBR:
Determining if the primitive arrays should be created is based on types supported by ParquetConverter serialization

NOTE: To ensure appending works as expected on parquets legacy schemas the first reader is created to obtain metadata, it adds small overhead for that case but ensures working out-of-the-box and limiting changes to ParquetSerializer


aloneguid:
Sorry for the delay, I couldn't figure out how to merge it in, but I have adopted your work in the latest version manually. Thanks a lot!

PR #352, opened 2023-06-18
Add support to .NET 6 TimeOnly type

ramon-garcia:
This pull request adds support to TimeOnly while trying to be as similar as possible to TimeSpan.

TimeOnly is more expressive when serializing data meant to be a time of the day rather than an interval of time. Too new fields are added to ParquetOptions to select TimeOnly instead of TimeSpan for deserializing: UseTimeOnlyTypeForTimeMillis (for time fields with millisecond precision) and UseTimeOnlyTypeForTimeMicros for selecting TimeOnly when one finds a time field with microsecond precision. This is similar to the filed UseDateOnlyTypeForDates for choosing DateOnly.

Best regards.


aloneguid (approved):
Perfect fit into existing codebase, thank you so much!

PR #363, opened 2023-06-29
Fix precision issues writing DateTime as milliseconds

### LINKED ISSUE #312: [BUG]: Loss of precision for DateTime milliseconds
### Library Version

4.8.1

### OS

Windows

### OS Architecture

64 bit

### How to reproduce?

1. Create a table with a `DateTimeDataField` using `DateTimeFormat.DateAndTime`
2. Add a row containing `new DateTime(9999, 12, 31, 23, 59, 59, 999, DateTimeKind.Utc)` (that is, 9999-12-31T23:59:59.999Z, or 253402300799999 milliseconds)
3. Write the table to a parquet file
4. Read the parquet file back and see that the value is 9999-12-31T23:59:59.998Z (i.e. off by one millisecond)

The issue comes down to the [ToUnixMilliseconds method](https://github.com/aloneguid/parquet-dotnet/blob/26ea729ac32de27b43f25ae08f9bb2f5a6efdfb9/src/Parquet/Extensions/OtherExtensions.cs#L28) using `TotalMilliseconds`, which is a double subject to loss of precision.

But there's another kind of rounding error if you use values prior to the epoch. For example, "1969-12-31T23:59:59.0001Z" (notice the fractional millisecond that should get truncated) round trips to "1969-12-31T23:59:59.001Z" (i.e. an extra millisecond).

For reference, check out the comments in [DateTimeOffset.ToUnixTimeSeconds](https://github.com/dotnet/runtime/blob/ab2b80d06dd4d997df5ffa72a3c4a99cb36ccbff/src/libraries/System.Private.CoreLib/src/System/DateTimeOffset.cs#L799) and their implementation of `ToUnixTimeMilliseconds` below it. Based on that, the most straightforward fix is to keep it as a long but also subtract _after_ the truncation to milliseconds:

```csharp
public static long ToUnixMilliseconds(this DateTime dto) {
    var milliseconds = dto.Ticks / TimeSpan.TicksPerMillisecond;
    return milliseconds - UnixEpoch.Ticks / TimeSpan.TicksPerMillisecond;
}
```

Though I've been wondering why use `DateTime` at all when `DateTimeOffset` both eliminates time zone ambiguity and provides `ToUnixTimeMilliseconds` and `FromUnixTimeMilliseconds` methods for you. As a middle ground, since the callers always ensure the `DateTime` is UTC before calling `ToUnixMilliseconds`, you could just convert to a `DateTimeOffset`: `return new DateTimeOffset(dto).ToUnixTimeMilliseconds();`.

### Failing test

```csharp
var data = new Table(new DateTimeDataField("d", DateTimeFormat.DateAndTime))
{
    new Row(new DateTime(9999, 12, 31, 23, 59, 59, 999, DateTimeKind.Utc)),
};
var file = Path.GetTempFileName();
await data.WriteAsync(file);
var result = await ParquetReader.ReadTableFromFileAsync(file);
Console.WriteLine(((DateTime)result[0][0]!).ToString("O"));
File.Delete(file);
```

spanglerco:
Fixes #312 by switching to [DateTimeOffset.ToUnixTimeMilliseconds](https://learn.microsoft.com/en-us/dotnet/api/system.datetimeoffset.tounixtimemilliseconds) instead of a custom implementation to keep things simple.

The first and third test cases fail without this change and all three pass with this change. I didn't know if there were any tests that actually write a parquet file and read it back that I should update. This PR just tests the method being modified.

### DIFF
--- a/src/Parquet.Test/Extensions/OtherExtensionsTest.cs
+++ b/src/Parquet.Test/Extensions/OtherExtensionsTest.cs
@@ -1,0 +1,20 @@
+﻿using System;
+using System.Globalization;
+using Xunit;
+
+namespace Parquet.Test.Extensions {
+    public class OtherExtensionsTest {
+
+        [Theory]
+        [InlineData("1969-12-31T23:59:59.0001Z", -1000)]
+        [InlineData("1970-01-01T00:00:00.0001Z", 0)]
+        [InlineData("9999-12-31T23:59:59.999Z", 253402300799999)]
+        public void ToUnixMilliseconds_truncates_milliseconds_correctly(string dateString, long expectedMilliseconds) {
+            var dateTime = DateTime.Parse(dateString, CultureInfo.InvariantCulture, DateTimeStyles.AdjustToUniversal);
+
+            long actualMilliseconds = dateTime.ToUnixMilliseconds();
+
+            Assert.Equal(expectedMilliseconds, actualMilliseconds);
+        }
+    }
+}
--- a/src/Parquet/Extensions/OtherExtensions.cs
+++ b/src/Parquet/Extensions/OtherExtensions.cs
@@ -6,6 +6,7 @@ using Parquet.Schema;
 namespace Parquet {
     static class OtherExtensions {
         private static readonly DateTime UnixEpoch = new DateTime(1970, 1, 1, 0, 0, 0, DateTimeKind.Utc);
+        private const long UnixEpochMilliseconds = 62_135_596_800_000L;
 
         public static int GetBitWidth(this int value) {
             for(int i = 0; i < 64; i++) {
@@ -26,7 +27,8 @@ namespace Parquet {
         }
 
         public static long ToUnixMilliseconds(this DateTime dto) {
-            return new DateTimeOffset(dto).ToUnixTimeMilliseconds();
+            long milliseconds = dto.Ticks / TimeSpan.TicksPerMillisecond;
+            return milliseconds - UnixEpochMilliseconds;
         }
 
         public static DateTime AsUnixDaysInDateTime(this int unixDays) {
--- a/src/Parquet/Extensions/OtherExtensions.cs
+++ b/src/Parquet/Extensions/OtherExtensions.cs
@@ -26,8 +26,7 @@ namespace Parquet {
         }
 
         public static long ToUnixMilliseconds(this DateTime dto) {
-            TimeSpan diff = dto - UnixEpoch;
-            return (long)diff.TotalMilliseconds;
+            return new DateTimeOffset(dto).ToUnixTimeMilliseconds();
         }
 
         public static DateTime AsUnixDaysInDateTime(this int unixDays) {


aloneguid:
Thanks a lot for moving it forward. I have no issues with it other than it being extremely slow (see below). Creating a new DTO object is expensive for a quick conversion. I'd rather prefer to extract the idea from .net source code (see https://github.com/dotnet/runtime/blob/299a8c9c178c38c6e7ca62b1bcce8f6e0d895ebe/src/libraries/System.Private.CoreLib/src/System/DateTimeOffset.cs#L799)

I've ran a simple performance test to prove the idea:

```csharp
[MeanColumn]
[MemoryDiagnoser]
[MarkdownExporter]
public class ArrayOps {

    private static readonly DateTime UnixEpoch = new DateTime(1970, 1, 1, 0, 0, 0, DateTimeKind.Utc);
    private static readonly DateTime dt = DateTime.UtcNow;
    const long UnixEpochTicks = 62135596800000L;
    
    [Benchmark(Baseline = true)]
    public void Old() {
        TimeSpan diff = dt - UnixEpoch;
        long epoch = (long)diff.TotalMilliseconds;
    }
    
    [Benchmark]
    public void New() {
        long epoch = new DateTimeOffset(dt).ToUnixTimeMilliseconds();
    }

    [Benchmark]
    public void NewOptimal() {
        long ms = dt.Ticks / TimeSpan.TicksPerMillisecond;
        long val = ms - UnixEpochTicks;
    }
}
```
results:

```
BenchmarkDotNet=v0.13.5, OS=Windows 11 (10.0.22621.1928/22H2/2022Update/SunValley2)

Intel Core i7-1065G7 CPU 1.30GHz, 1 CPU, 8 logical and 4 physical cores

.NET SDK=7.0.304

  [Host] : .NET 7.0.7 (7.0.723.27404), X64 RyuJIT AVX2


|     Method |       Mean |     Error |    StdDev |     Median | Ratio | RatioSD | Allocated | Alloc Ratio |
|----------- |-----------:|----------:|----------:|-----------:|------:|--------:|----------:|------------:|
|        Old |  0.0054 ns | 0.0073 ns | 0.0068 ns |  0.0022 ns |     ? |       ? |         - |           ? |
|        New | 12.9621 ns | 0.1942 ns | 0.1816 ns | 12.9618 ns |     ? |       ? |         - |           ? |
| NewOptimal |  0.0078 ns | 0.0147 ns | 0.0157 ns |  0.0000 ns |     ? |       ? |         - |           ? |
```

3rd option is about 1661 times faster than the second.

What do you think?

spanglerco:
The more optimal implementation looks good to me, thanks.

plaisted:
FYI your benchmark code has a bug I think, `long val = ms - UnixEpochTicks` should be `long val = ms - UnixEpochMilliseconds`. Won't affect performance but thought I'd bring it up if it gets copied to real code base.

spanglerco:
Yes, the version that I included in the PR uses `UnixEpochMilliseconds`.

@aloneguid let me know if there are any other changes you want me to make.

aloneguid (approved):
sorry for the delay, looks good to me

aloneguid:
will be released in 4.16.0

Issue #361, opened 2023-06-29
Does the library support treating missing columns as nulls?


greenlynx:
### Issue description

Hi,

I have a Parquet file that I wrote some records to using `ParquetSerializer.SerializeAsync<MyRecordType>`. I'm using `ParquetSerializer.DeserializeAsync<MyRecordType>` to read it back.

I'd now like to add a new, nullable string property named `AccountId` to `MyRecordType`, but retain the ability to read the existing file. I'm getting an exception: `Parquet.ParquetException: 'AccountId' does not exist in this file` which makes sense, but it would be good to have the option to just set the new property to be set to null when the column doesn't exist in the file. Is that possible somehow please?

aloneguid:
At the moment you can mark this new property with `[JsonIgnore]` and deserializer will skip it. Note that schema evolution is not supported by parquet format itself.

greenlynx:
Thanks for the fast reply - as I understand it, `[JsonIgnore]` also stops the field being serialized too, which isn't what I need. I need to read the file (with the new field set to null if it's missing from the Parquet file) then write it again, including the new field.

Would you be open to a PR that enables this "treat missing columns as null" behavior (opt-in, behind an option?)

greenlynx:
Hi - I've submitted a PR to address this functionality: https://github.com/aloneguid/parquet-dotnet/pull/366

PR #366, opened 2023-06-30
Add SkipDeserializingMissingColumns option to allow handling missing fields in files gracefully

greenlynx:
Addresses #361

### DIFF
--- a/src/Parquet.Test/Serialisation/ParquetSerializerTest.cs
+++ b/src/Parquet.Test/Serialisation/ParquetSerializerTest.cs
@@ -41,6 +41,54 @@ namespace Parquet.Test.Serialisation {
             Assert.Equivalent(data2, data);
         }
 
+        class RecordWithNewField : Record {
+            public DateTime NewTimestamp { get; set; }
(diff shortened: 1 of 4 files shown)


aloneguid (commented):
It looks good, thank you. I'm just wandering if it would make sense to make this a default behavior rather than opt-in. I think skipping by default is magic, and exceptions like before are not ;)

aloneguid (approved):
looks good to me, thanks. I'll update the docs before release.

aloneguid:
Actually I like it too much to postpone, it's not a breaking change, just avoids more support requests. Thanks again ;)

tukaef:
Hello, has this improvement been released in 4.16.0? I cannot find `SkipDeserializingMissingColumns` in `ParquetOptions`

eFse7en:
> There's no option. It just magically works.

Why was this implemented as default behavior with no option to force the check?

I've just spent 2 hours debugging nulls in a column before I've realized that the new parquet file had a casing changed in a column. So the actual column was ignored without any error.

PR #368, opened 2023-07-10
Feature 252 expose column statistics before data load

mirosuav:
Hi Aloneguid.

Finally I've got a while to take care of this feature.

To get access to the data column statistics before loading the column data I've made two changes to the code:

1.  Make the  _ParquetRowGroupReader_ exposing _DataColumnReader_
with new public method: `DataColumnReader GetColumnReader(DataField field)`

2.  Make the _DataColumnReader_ part of public api and exposing statistics with method 
`DataColumnStatistics? GetColumnStatistics()`

Thanks
Mirek


src/Parquet/ParquetRowGroupReader.cs
```diff
@@ -86,6 +80,24 @@ public class ParquetRowGroupReader : IDisposable {
...
         }
 
+        /// <summary>
+        /// Returns DataColumnReader for given field
+        /// </summary>
+        /// <param name="field">DataField</param>
+        /// <returns></returns>
+        /// <exception cref="ArgumentNullException"></exception>
+        /// <exception cref="ParquetException"></exception>
+        public DataColumnReader GetColumnReader(DataField field) {
```
aloneguid, line 90: It looks good to me overall, thank you!
I'm more thinking in terms of keeping API consistent, above this method there is `GetMetadata()` and `GetCustomMetadata()` methods, and your addition would fit perfectly well as another method like `GetStatistics(DataField)`?

In this case one doesn't need to open `DataColumnReader` on the client, and you can still keep the `DataColumnReader` class private to avoid supporting it in public API?
```suggestion
        public DataColumnReader GetColumnReader(DataField field) {
```

aloneguid:
Moved to #383

PR #373, opened 2023-07-25
Use RecycableMemoryStream in DataColumnWriter

itayfisz:
In DataColumnWriter, RecycableMemoryStream wasn't used in a particular case, and instead MemoryStream was initialized directly.


aloneguid (approved):
looks good, thank you!

Issue #375, opened 2023-08-01
Add DELTA_BINARY_PACKED encoder For Int32 And Int64


ee-naveen:
As of now, parquet.net does not support delta encoding.

https://github.com/aloneguid/parquet-dotnet/issues/205 Parquet.net already supports read, however write is currently not supported.

Delta Binary Packed encoding is designed to achieve high compression ratios for numerical data types. 
It efficiently packs the numerical values, reducing the amount of storage required to store the data on disk. 
This can result in significant storage space savings, especially for datasets with large numeric columns.

Reference:

- https://image.slidesharecdn.com/hadoopsummit-140630160016-phpapp01/75/efficient-data-storage-for-analytics-with-apache-parquet-20-29-2048.jpg?cb=1665769903
- https://image.slidesharecdn.com/hadoopsummit-140630160016-phpapp01/75/efficient-data-storage-for-analytics-with-apache-parquet-20-30-2048.jpg?cb=1665769903
- https://image.slidesharecdn.com/hadoopsummit-140630160016-phpapp01/75/efficient-data-storage-for-analytics-with-apache-parquet-20-31-2048.jpg?cb=1665769903
- https://www.slideshare.net/cloudera/hadoop-summit-36479635

aloneguid:
Thanks, this is definitely on my radar.

ee-naveen:
I have made significant progress, and we might be raising a PR for review soon.
The initial results look promising, but running more tests.

ee-naveen:
Available in 4.16

PR #382, opened 2023-08-08
Add DELTA_BINARY_PACKED encoder For Int32 and Int64

ee-naveen:
Add DELTA_BINARY_PACKED encoder For Int32 and Int64 #375 
- Added Column Encoding in ParquetOptions class
- Added DeltaBinaryPackedEncoder.Encode.Variations.tt file for encoding variation

### DIFF
--- a/src/Parquet/Encodings/DeltaBinaryPackedEncoder.Encode.Variations.tt
+++ b/src/Parquet/Encodings/DeltaBinaryPackedEncoder.Encode.Variations.tt
@@ -12,8 +12,6 @@
     //https://github.com/xitongsys/parquet-go/blob/62cf52a8dad4f8b729e6c38809f091cd134c3749/encoding/encodingwrite.go#L287
 
     static partial class DeltaBinaryPackedEncoder {
-         private static readonly ArrayPool<byte> BytePool = ArrayPool<byte>.Shared;
-
 <# foreach(string nt in types) { #>
         private static void Encode(ReadOnlySpan<<#= nt#>> data, Stream destination) {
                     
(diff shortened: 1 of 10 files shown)


eeabhishekmore:
@aloneguid  please take a look when you can, thanks

aloneguid:
Will do. I'll try to include this in the upcoming release.

src/Parquet/ParquetOptions.cs
```diff
@@ -47,5 +48,14 @@ public class ParquetOptions {
...
         public double DictionaryEncodingThreshold { get; set; } = 0.8;
+
+        /// <summary>
+        /// Gets or sets the dictionary that stores column encodings for different data fields.
+        /// </summary>
+        /// <remarks>
+        /// The dictionary maps column names (as strings) to their corresponding encodings.
+        /// Each column can be encoded using a specific encoding method defined by the 'Encoding' enum.
+        /// </remarks>
+        public Dictionary<string, Encoding> ColumnEncoding { get; set; } = new Dictionary<string, Encoding>();
```
aloneguid, line 59: How do we use this?
Are only primitive column names at root supported as a key?
Are all `Encoding`s respected?

src/Parquet/Encodings/DeltaBinaryPackedEncoder.cs
```diff
@@ -9,24 +12,120 @@ namespace Parquet.Encodings {
     /// 
     /// Supported Types: INT32, INT64
     /// </summary>
-    static partial class DeltaBinaryPackedEncoder {
+    public static partial class DeltaBinaryPackedEncoder {
```
aloneguid, line 15: should stay private

src/Parquet/Encodings/DeltaBinaryPackedEncoder.Encode.Variations.tt
```diff
@@ -0,0 +1,93 @@
...
+            int numValuesInMiniBlock = 32;
+            int totalNumValues = data.Length;
+
+            WriteUnsignedVarInt(destination, blockSize);
+            WriteUnsignedVarInt(destination, numMiniBlocksInBlock);
+            WriteUnsignedVarInt(destination, totalNumValues);
+
+            <#= nt#> num = data[0];
+            ulong firstValue = ZigZagEncode(num);
+            WriteUnsignedVarInt(destination, firstValue);
```
aloneguid, line 30: In case of long, will this lose precision?

src/Parquet/Encodings/DeltaBinaryPackedEncoder.cs
```diff
@@ -9,24 +12,120 @@ namespace Parquet.Encodings {
...
+            } else if(t == typeof(long[])) {
+                Encode((long[])data, destination);
+                if(stats != null)
+                    ParquetPlainEncoder.FillStats((long[])data, stats);
+            } else {
+                throw new NotSupportedException($"only {Parquet.Meta.Type.INT32} and {Parquet.Meta.Type.INT64} are supported in {Encoding.DELTA_BINARY_PACKED} but type passed is {t}");
+            }
+        }
+
+        private static ulong ZigZagEncode(long num) {
```
aloneguid, line 71: should be moved to OtherExtensions.cs

src/Parquet/Encodings/DeltaBinaryPackedEncoder.cs
```diff
@@ -9,24 +12,120 @@ namespace Parquet.Encodings {
...
+            int consumed = 0;
+            try {
+                WriteUnsignedVarInt(rentedBuffer, ref consumed, value);
+                destination.Write(rentedBuffer, 0, consumed);
+            } finally {
+                BytePool.Return(rentedBuffer);
+            }
+        }
+        private static void WriteUnsignedVarInt(Stream destination, ulong value) {
+            byte[] rentedBuffer = BytePool.Rent(8);
```
aloneguid, line 90: FYI renting small buffers is expensive, probably more than creating them directly.

src/Parquet/ParquetOptions.cs
ee-naveen: *How do we use this?*
Here is the sample.
```
...
var dataField = new DataField<long>("column1");
ParquetOptions options = new ParquetOptions() {
  ColumnEncoding = new Dictionary<string, Meta.Encoding>() {
      	{ "column1", Meta.Encoding.DELTA_BINARY_PACKED }
     }
};
...
```

*Are all Encodings respected?*

Currently only DELTA_BINARY_PACKED is supported for column encoding.
Here is the validation function
```
        private static bool EnsureColumnEncodingIsSupported(Encoding columnEncoding) {
            switch(columnEncoding) {
                case Encoding.DELTA_BINARY_PACKED:
                    return true;
                default:
                    throw new ArgumentException($"Not supported column encoding {columnEncoding}");
            }
        }
```

src/Parquet/Encodings/DeltaBinaryPackedEncoder.Encode.Variations.tt
ee-naveen: added WriteZigZagVarLong function and refactored (removed other functions)

src/Parquet/Encodings/DeltaBinaryPackedEncoder.cs
ee-naveen: Thank you for the information. I've updated the approach and now creating small buffers directly.

src/Parquet/Encodings/DeltaBinaryPackedEncoder.cs
ee-naveen: Added GetZigZagEncoded extension method for long in OtherExtensions.cs

src/Parquet/Encodings/DeltaBinaryPackedEncoder.cs
ee-naveen: Reverted to private.

aloneguid:
Apologies this is so slow. During the review I've realised the other parts of this lib have a lot of duplications around variable-length and zigzag encoding methods. So the easiest way to do that for me was to copy this code in another branch and continue working on it. I've copied most of the code already.

The other thing about configuration I'm not sure about is whether it needs to be configurable. For instance, `delta-rs` project just uses delta binary packed encoding by default for int32 and int64, so it might be also worth making your code the default for writing to make things simple for a user. Thoughts?

aloneguid:
Closing this in favor of merging into #385. Thanks again!

ee-naveen:
Thank you @aloneguid for picking this up.

Our current use case is fully compatible with the default delta binary packed encoding for integers.
If necessary, we can consider adding support for configuration in the future.

PR #388, opened 2023-08-18
Use PLAIN encoding for columns without defined data

spanglerco:
Part of #375. If a column only contains null values, the array passed to `DeltaBinaryPackedEncoder` will be empty and cause the encoder to return without writing a header. This breaks compatibility with some readers such as Apache parquet-mr that try to read the header.

I verified writing a parquet file containing such a column can be read by parquet-mr with this change but failed before.


aloneguid (approved):
lgtm, thanks a lot, really appreciate it!

Issue #389, opened 2023-08-18
[BUG]: ParquetRowGroupReader.ReadColumnAsync returning wrong values for Int32 columns


El-Gor-do:
### Library Version

4.16.0

### OS

Windows

### OS Architecture

64 bit

### How to reproduce?

TestParquet.csproj
```
<Project Sdk="Microsoft.NET.Sdk">
  <PropertyGroup>
    <OutputType>Exe</OutputType>
    <TargetFramework>net7.0</TargetFramework>
  </PropertyGroup>
  <ItemGroup>
    <PackageReference Include="Parquet.Net" Version="4.15.0" />
  </ItemGroup>
</Project>
```

Program.cs
```
using Parquet;
using Parquet.Data;
using Parquet.Schema;
using Parquet.Serialization;
using System;
using System.Collections.Generic;
using System.IO;
using System.IO.Compression;
using System.Linq;
using System.Threading.Tasks;

namespace TestParquet
{
    internal class Program
    {
        public class TestClass
        {
            public int Value { get; set; }

            private static Random r { get; } = new Random(0);

            public TestClass()
            {
                this.Value = TestClass.r.Next(int.MinValue, int.MaxValue);
            }
        }

        static async Task Main(string[] args)
        {
            // create items
            int itemCount = 4;
            List<TestClass> items = Enumerable.Range(0, itemCount).Select(i => new TestClass()).ToList();

            List<int> actualValues = new List<int>();
            using (MemoryStream ms = new MemoryStream())
            {
                // create parquet stream from items
                ParquetSerializerOptions options = new ParquetSerializerOptions()
                {
                    Append = false,
                    CompressionLevel = CompressionLevel.SmallestSize,
                    CompressionMethod = CompressionMethod.Gzip,
                };
                ParquetSchema schema = await ParquetSerializer.SerializeAsync(items, ms, options);
                ms.Position = 0;

                // read values in parquet stream
                DataField field = schema.DataFields[0];
                ParquetReader reader = await ParquetReader.CreateAsync(ms, leaveStreamOpen: true);
                for (int rowGroupIndex = 0; rowGroupIndex < reader.RowGroupCount; ++rowGroupIndex)
                {
                    using (ParquetRowGroupReader rowGroupReader = reader.OpenRowGroupReader(rowGroupIndex))
                    {
                        // if itemCount > 4096 then this throws InvalidOperationException: 'don't know how to skip'
[...]

PR #391, opened 2023-08-22
[DELTA_BINARY_PACKED Decoding]: Adding first value to destination array before reading the block

ee-naveen:
[DELTA_BINARY_PACKED Decoding]: Adding first value to destination array before reading the block
#389


aloneguid (approved):
awesome, thank you

PR #423, opened 2023-10-27
Allow deserialization from open RowGroupReaders.

### LINKED ISSUE #422: ParquetSerializer should allow deserialization of open RowGroupReaders
### Issue description

The current implementation supports `ParquetSerializer.DeserializeAsync` with a `rowGroupIndex`, but iterating over many of these would open and close a `ParquetReader` every time.  An additional method should be available that reads directly from a `RowGroupReader` so that only a single `ParquetReader` is needed.

ddrinka:
Add a new public method that matches the existing deserialization functionality but operates on open `RowGroupReaders` rather than opening and closing a `ParquetReader` for every row group.

Closes #422

### DIFF
--- a/src/Parquet/Serialization/ParquetSerializer.cs
+++ b/src/Parquet/Serialization/ParquetSerializer.cs
@@ -155,6 +155,30 @@ namespace Parquet.Serialization {
             return result;
         }
 
+        /// <summary>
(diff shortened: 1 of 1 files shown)


aloneguid (approved):
Thanks, this can be potentially very useful

Issue #407, opened 2023-09-25
[BUG]: Performance decrease since > 4.11 or higher when writing into Azure blob stream


andreaslennartz:
### Library Version

4.11, 4.12, 4.13, 4.14, 4.15, 4.16

### OS

Windows 11,Azure, .NET 6

### OS Architecture

64 bit

### How to reproduce?

First of all, thanks for creating this great library! 
We encountered an issue when using Parquet.NET to directly write data using a stream on an Azure blob. 

**Bug details:**
Parquet.NET supports writing into existing stream (e.g. File stream). We write data into an Azure blob using the nuget Package `Azure.Storage.Blobs` and a `BlockBlobClient`. 
Here is an example how the writer is initialized: 

```C#
var blockBlobClient = new BlockBlobClient(connectionString, containerName, name);
using var stream = blockBlobClient.OpenWrite(true, null);
using StreamWriter streamWriter = new StreamWriter(stream);
using var parquetWriter = await ParquetWriter.CreateAsync(schema, streamWriter.BaseStream, null, append: false);
```

Since we upgraded from Parquet.NET version 4.10.1 to a higher version (e.g. the current 4.16.4, but all other version higher than 4.11 are also affected), the time of writing into the blob stream decreased significantly (about 10x to 30x times slower). 

Please see the attached example project that shows how to reproduce the issue (an Azure storage account is needed to run the test, please reach out to me if don't have access to one, I can share a temp account for testing). 

**How to reproduce the issue:  **
1. Start the example project
2. Set the connection string to Azure Blob Storage (contact me if you don't have access to an Azure storage account)
3. Create a container with the same name as the variable 'containerName'
3. Run the code with Parquet.NET 4.11.X or higher, measure exeuction time
4. Downgrade package reference to Parquet.NET 4.10.1
5. Run the code again, compare execution time with previous run

Result in my environment: 
Upload time using Parquet.NET 4.10.1: < 10 sec
Upload time using Parquet.NET > 4.11.X or higher: > 150 sec

Example project:
[AzureBlobStorageFlushIssue.zip](https://github.com/aloneguid/parquet-dotnet/files/12713537/AzureBlobStorageFlushIssue.zip)

**Proposed solution**
(Please note that these are only ideas which I want to share, please feel free to comment on them!)

[...]

PR #432, opened 2023-11-19
#407 added option to reduce flushing of streams during write operation.

dxdjgl:


src/Parquet/ParquetOptions.cs
```diff
@@ -6,6 +6,12 @@ namespace Parquet {
     /// Parquet options
     /// </summary>
     public class ParquetOptions {
+
+        /// <summary>
+        /// Reduce flushing of stream during write, to increase performance of for instance Azure blob
+        /// </summary>
+        public bool ReduceStreamFlushing { get; set; }
```
aloneguid, line 13: I think it might be better to reduce flushing by default and have no option - less permutations to support.

aloneguid: i took the liberty of doing so, hope you don't mind

dxdjgl: Sorry but I dont agree, since it was found out that removing that flush will make parquet.net hang in this method call
        public Task WriteColumnAsync(DataColumn column, CancellationToken cancellationToken = default)
in some schenarios. So I still believe the best solution for now is to have the flush handling configurable, until a permanent fix for the stuck Task has been found.

aloneguid: Sorry missed this issue. I'll add it back now and release a patch. Is there any way we can reproduce the failure with a unit test btw?

dxdjgl: Please have a look at the code I sent in a mail to you.

PR #433, opened 2023-11-19
Add DeserializeAllAsync

### LINKED ISSUE #386: IAsyncEnumerable deserializer
https://github.com/aloneguid/parquet-dotnet/blob/07e7c6792a08c5e227819db65d2483c99a58e2f5/src/Parquet/Serialization/ParquetSerializer.cs#L140

There should be a version of this method that returns an `IAsyncEnumerable<T>`. Implementation like this, maybe? (Happy to make a PR if desired.)

```csharp
Assembler<T> asm = GetAssembler<T>();
var result = new List<T>();

using ParquetReader reader = await ParquetReader.CreateAsync(source, options, cancellationToken: cancellationToken);
for(int rgi = 0; rgi < reader.RowGroupCount; rgi++) {
     await DeserializeRowGroupAsync(reader, rgi, asm, result, cancellationToken);
     foreach (var item in result) { 
        yield return item; 
    }
    result.Clear();
}
```

Arithmomaniac:
Fixes #386 .

### DIFF
--- a/src/Parquet.Test/Parquet.Test.csproj
+++ b/src/Parquet.Test/Parquet.Test.csproj
@@ -18,6 +18,7 @@
 
     <ItemGroup>
         <PackageReference Include="Newtonsoft.Json" Version="13.0.3" />
+        <PackageReference Include="System.Linq.Async" Version="6.0.1" />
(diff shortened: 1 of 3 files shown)


aloneguid (approved):
lgtm, thank you

Issue #133, opened 2021-11-12
Question: How would I specify CultureInfo.InvariantCulture alongside table.ToString("j")


RushkaVenter:
**Version:** Parquet.Net v3.9.1

**Runtime Version:** .Net Core 3.1

**OS:** Windows/Linux/MacOSX v etc.

#### Expected behavior

When using ToString("j") on a Parquet.Data.Rows.Row object, the string should return a string formatted like a json string.

#### Actual behavior

When using ToString("j") on a Parquet.Data.Rows.Row object, the string returns a string formatted like a json string, however, the decimal values within the parquet is "12,5" and the json string looks like {"property_name": 12,5, "property_name_2: value} which is not a valid json string.

Normally you would be able to specify CultureInfo.InvariantCulture to get the decimals to convert as "12.5". How would you do that here?

### Steps to reproduce the behavior

1. Create an ParquetReader from a file with decimal values
2. Read the parquet file as a table.
3. Select the first row of the table
4. Convert the the row to string
5. Deserialize the json string

#### Code snippet reproducing the behavior

```csharp
 using (var parquetReader = new ParquetReader(ms))
  {
      var table = parquetReader.ReadAsTable();
      var firstItem = table.FirstOrDefault();
      Console.WriteLine(firstItem.ToString("j"));
      var obj = JsonSerializer.Deserialize<ItemModel>(firstItem.ToString("j"));
  }
```

aloneguid:
this is by design, you're welcome to improve it

aloneguid:
Reopening it. Feel free to raise a PR, contributions are always welcome!

PR #438, opened 2023-12-07
Fix/encode json numbers with invariant culture

rachied:
Explicitly use invariant culture when encoding number types, eliminating the potential for generating invalid JSON.

Fixes https://github.com/aloneguid/parquet-dotnet/issues/133


aloneguid (approved):
looks good to me, thank you!

Issue #437, opened 2023-12-02
[BUG]: ParquetSerializer.DeserializeAsync - "Not a parquet" file when using buffer with fixed size


edguer:
### Library Version

4.17.0

### OS

Windows

### OS Architecture

64 bit

### How to reproduce?

Write this code or similar:

```CSharp
var bytes = [some byte array];

// Using a 4096 buffer from the array pool
using var memoryStream = new MemoryStream(ArrayPool<byte>.Shared.Rent(4096));
await memoryStream.WriteAsync(bytes, 0, bytes.Length);
return await ParquetSerializer.DeserializeAsync<T>(memoryStream, ParquetOptions.ParquetOptions);
```

And you get:
```
  System.IO.IOException: not a parquet file, head: 50415231, tail: 00000000
Stack Trace: 
  ParquetActor.ValidateFileAsync()
  ParquetReader.InitialiseAsync(CancellationToken cancellationToken)
  ParquetReader.CreateAsync(Stream input, ParquetOptions parquetOptions, Boolean leaveStreamOpen, CancellationToken cancellationToken)
  ParquetSerializer.DeserializeAsync[T](Stream source, ParquetOptions options, CancellationToken cancellationToken)
```

The reason is the ValidateFileAsync method seeks the tail at the end of stream, which will always be 0 when the buffer is not filled:

https://github.com/aloneguid/parquet-dotnet/blob/9dcd02ec3e2120d31dd85ed31afc9255c2fc8f5c/src/Parquet/ParquetActor.cs#L42C50-L42C50
```CSharp
        protected async Task ValidateFileAsync() {
            _fileStream.Seek(0, SeekOrigin.Begin);
            byte[] head = await _fileStream.ReadBytesExactlyAsync(4);

            _fileStream.Seek(-4, SeekOrigin.End);
            byte[] tail = await _fileStream.ReadBytesExactlyAsync(4);

            if(!MagicBytes.SequenceEqual(head) || !MagicBytes.SequenceEqual(tail))
                throw new IOException($"not a parquet file, head: {head.ToHexString()}, tail: {tail.ToHexString()}");
        }
```

This would probably fix it:

```CSharp
            _fileStream.Seek((_fileStream.Position - 4), SeekOrigin.Begin);
            byte[] tail = await _fileStream.ReadBytesExactlyAsync(4);
```

aloneguid:
Sounds good. Please raise a PR if that's ok.

daz10000:
[...]

PR #440, opened 2023-12-16
Update ParquetActor.cs

fasterinnerlooper:
Address issue found in #437

### DIFF
--- a/src/Parquet/ParquetActor.cs
+++ b/src/Parquet/ParquetActor.cs
@@ -39,7 +39,7 @@ namespace Parquet {
             _fileStream.Seek(0, SeekOrigin.Begin);
             byte[] head = await _fileStream.ReadBytesExactlyAsync(4);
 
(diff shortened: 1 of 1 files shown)


src/Parquet/ParquetActor.cs
```diff
@@ -39,7 +39,7 @@ public class ParquetActor {
             _fileStream.Seek(0, SeekOrigin.Begin);
             byte[] head = await _fileStream.ReadBytesExactlyAsync(4);
 
-            _fileStream.Seek(-4, SeekOrigin.End);
+            _fileStream.Seek((_fileStream.Position - 4), SeekOrigin.Begin);
```
aloneguid, line 42: Sorry just trying to figure out why this will work. Say we have a stream of 10 bytes.
- after line 40, it's at position 4.
- seek goes to 4 - 4 from beginning, meaning it goes to position 0.

That's not right, as it should be at position 10 - 4?

Issue #461, opened 2024-01-20
[BUG]: Append serialize to file-path actually truncates file and then throws because it's empty


danielearwicker:
### Library Version

4.23.1

### OS

All

### OS Architecture

64 bit

### How to reproduce?

Similar to the example in the docs:

```cs
await ParquetSerializer.SerializeAsync(data, filePath, new ParquetSerializerOptions { Append = false });
await ParquetSerializer.SerializeAsync(data, filePath, new ParquetSerializerOptions { Append = true });
```

`SerializeAsync` uses `File.Create(filePath)` which truncates the file, regardless of the `Append` flag.

### Failing test

```csharp
try
{
    await ParquetSerializer.SerializeAsync(
        new[]
        {
            new Record 
            {
                Timestamp = DateTime.UtcNow,
                EventName = "first",
                MeterValue = 1
            }
        }, 
        tempPath, 
        new ParquetSerializerOptions { Append = false });
    
    await ParquetSerializer.SerializeAsync(
        new[]
        {
            new Record 
            {
                Timestamp = DateTime.UtcNow,
                EventName = "second",
                MeterValue = 2
            }
        }, 
        tempPath,
        new ParquetSerializerOptions { Append = true });
    
    using ParquetReader reader = await ParquetReader.CreateAsync(tempPath);
    
    using (ParquetRowGroupReader reader0 = reader.OpenRowGroupReader(0))
    {
        Assert.Equal(1, reader0.RowCount);
    }
    
    using (ParquetRowGroupReader reader1 = reader.OpenRowGroupReader(1))
    {
        Assert.Equal(1, reader1.RowCount);
    }
}
finally 
{
    if (tempPath != null)
    {
        System.IO.File.Delete(tempPath);
    }
}
```

aloneguid:
merged, please give it a go ;)

PR #462, opened 2024-01-20
#461 Avoid file truncation when serializing with Append = true

danielearwicker:
Awkward having to create a temp file in a unit test but have taken care that the file will be deleted.


aloneguid:
thanks, that looks good

aloneguid (approved):
lgtm

aloneguid (approved):

Issue #465, opened 2024-01-24
[BUG]: SerializeAsync throws when serializing nullable decimal


stefer:
### Library Version

4.23.0

### OS

Windows/Linux

### OS Architecture

64 bit

### How to reproduce?

The following code throws exception when version 4.23.0 or larger is used.

Exception:
```
Expression of type 'System.Nullable`1[System.Decimal]' cannot be used for parameter of type 'System.Decimal' of method 'Void Add(System.Decimal)' (Parameter 'arg0')
```

Stacktrace:
```
   at System.Dynamic.Utils.ExpressionUtils.ValidateOneArgument(MethodBase method, ExpressionType nodeKind, Expression arguments, ParameterInfo pi, String methodParamName, String argumentParamName, Int32 index)
   at System.Linq.Expressions.Expression.Call(Expression instance, MethodInfo method, Expression arg0)
   at System.Linq.Expressions.Expression.Call(Expression instance, MethodInfo method, IEnumerable`1 arguments)
   at Parquet.Serialization.Dremel.FieldStriperCompiler`1.WriteValue(ParameterExpression valueVar, Int32 dl, Expression currentRlVar, ParameterExpression isLeaf, Boolean isAtomic)
   at Parquet.Serialization.Dremel.FieldStriperCompiler`1.WhileBody(Expression element, Boolean isAtomic, Int32 dl, ParameterExpression currentRlVar, ParameterExpression seenFieldsVar, Field field, Int32 rlDepth, Type elementType, List`1 path)
   at Parquet.Serialization.Dremel.FieldStriperCompiler`1.DissectRecord(Expression rootVar, Field parentField, Field[] levelFields, List`1 path, Type rootType, Int32 rlDepth, ParameterExpression currentRlVar)
   at Parquet.Serialization.Dremel.FieldStriperCompiler`1.Compile()
   at Parquet.Serialization.Dremel.Striper`1.CreateStriper(DataField df)
   at System.Linq.Enumerable.SelectArrayIterator`2.Fill(ReadOnlySpan`1 source, Span`1 destination, Func`2 func)
   at System.Linq.Enumerable.SelectArrayIterator`2.ToList()
   at Parquet.Serialization.Dremel.Striper`1..ctor(ParquetSchema schema)
   at Parquet.Serialization.ParquetSerializer.<>c__6`1.<SerializeAsync>b__6_0(Type _)
[...]

PR #468, opened 2024-01-26
Serialize nullable nullable types (#465)

stefer:
Attempt to detect nullable for any type.
I am not sure how this relates to IsRequired and I have not added deserialization tests.
Maybe tests should be added for all data types - I just added a few.

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -8,3 +8,4 @@ BenchmarkDotNet.Artifacts/
 Parquet.sln.DotSettings.user
 launchSettings.json
 .DS_Store
+*.user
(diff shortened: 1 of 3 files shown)


stefer:
Hmm, yea, I could not run all unit tests locally (java references), so I missed these.
I also think my solution was rather naive - we should probably look into the changes that where made for fields and see why that introduced regression. I need some more time if I should dig deeper.

aloneguid:
Hey, I appreciate your hard work on this, it really helped me out. I made the changes you suggested and pushed them to the new version.

PR #485, opened 2024-03-08
Bug: Add failing unit test for delta binary packed deserialisation issue

swindiggie:
This adds a failing unit test to highlight an issue with deserialising delta binary packed encoded data. Hopefully this can be used to help find a solution.

Github Issue:
https://github.com/aloneguid/parquet-dotnet/issues/486


swindiggie:
Closing because this test was added in https://github.com/aloneguid/parquet-dotnet/pull/487 🙌

Issue #486, opened 2024-03-08
[BUG]: Deserialising delta binary packed encoded data produces incorrect results


swindiggie:
### Library Version

4.23.4

### OS

MacOS 13.4

### OS Architecture

ARM 64

### How to reproduce?

I generated a parquet file using the following one-column CSV:
```
"column1"
1000
1
2
3
4
5
6
7
8
9
10
```

The single column in the parquet file is Int32 using DELTA_BINARY_PACKED encoding.

When I deserialise the parquet file using parquet-dotnet, I get:
```
"column1"
1000
1
-998
-1997
-2996
-3995
-4994
-5993
-6992
-7991
-8990
```

The incorrect results can be viewed using Parquet Floor 4.23.4.

I have also opened a PR with a failing unit test against the parquet file as a reference:
 - https://github.com/aloneguid/parquet-dotnet/pull/485

I verified a few other websites display correct results for the parquet file:
 - https://parquetreader.com
 - https://www.parquet-viewer.com
 - https://konbert.com/viewer/parquet

The CSV file and a screenshot of Parquet Floor are attached.

[test.csv](https://github.com/aloneguid/parquet-dotnet/files/14539931/test.csv)

<img width="288" alt="Screenshot 2024-03-07 at 4 26 59 pm" src="https://github.com/aloneguid/parquet-dotnet/assets/31874382/cceb5ab0-87b1-4623-a68d-64ba624d1802">

PR #487, opened 2024-03-12
fix: attempt to fix issues where loop will read past the end of a block

alex-harper:
This PR is a fix for https://github.com/aloneguid/parquet-dotnet/issues/486

The delta binary packed encoder has a few bugs where it will read past the end of a block and fill the destination buffer with incorrect output.

This PR terminates the various loops once the final value for a block has been read.


aloneguid (approved):
LGTM

PR #490, opened 2024-03-19
Nullable Timespan

cliedeman:


src/Parquet/Serialization/TypeExtensions.cs
```diff
@@ -139,7 +139,7 @@ class ClassFieldMember : ClassMember {
                     member?.MicroSecondsTimeAttribute == null
                         ? TimeSpanFormat.MilliSeconds
                         : TimeSpanFormat.MicroSeconds,
-                    isNullable, null, propertyName);
```
cliedeman, line 142: The decorator should probably always override the property type

```
[ParquetRequired]
public TimeSpan? TimeSpan {get; init; }
```

not null

aloneguid (approved):
LGTM

PR #516, opened 2024-06-04
121: in DataColumn set DefinitionLevel to MaxDefinitionLevel - 1 if value is null.

greg0rym:
121: in DataColumn set DefinitionLevel to MaxDefinitionLevel - 1 if value is null.

### DIFF
--- a/src/Parquet.Test/Types/ListOfStructuresTest.cs
+++ b/src/Parquet.Test/Types/ListOfStructuresTest.cs
@@ -1,0 +1,75 @@
+﻿using System;
+using System.IO;
+using System.Numerics;
+using System.Threading.Tasks;
+using Parquet.Data;
+using Parquet.File.Values.Primitives;
+using Parquet.Rows;
+using Parquet.Schema;
+using Xunit;
+
+namespace Parquet.Test.Types {
+    public class ListOfStructuresTest : TestBase {
+        [Theory]
+        [MemberData(nameof(NullableDataFields))]
+        public async Task List_of_structures_with_nullable_fields_sets_correct_definition_levels(DataField nullableDataField) {
+
+            var schema = new ParquetSchema(
+                new DataField<int>("data_field"),
+                new ListField("list_field",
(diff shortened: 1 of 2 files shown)


src/Parquet.Test/Types/ListOfStructuresTest.cs
```diff
@@ -0,0 +1,75 @@
...
+            new DataField<bool?>("bool"),
+            new DataField<byte?>("byte"),
+            new DataField<sbyte?>("sbyte"),
+            new DataField<short?>("short"),
+            new DataField<ushort?>("ushort"),
+            new DataField<int?>("int"),
+            new DataField<uint?>("uint"),
+            new DataField<long?>("long"),
+            new DataField<ulong?>("ulong"),
+            //new DataField<BigInteger?>("BigInteger"),
```
greg0rym, line 61: I am not sure how to test BigInteger:

image

src/Parquet/Extensions/UntypedArrayExtensions.cs
```diff
@@ -458,7 +458,7 @@ static class UntypedArrayExtensions {
                 bool? value = array[i];
 
                 if(value == null) {
-                    dest[y] = 0;
+                    dest[y] = fillerValue - 1;
```
aloneguid, line 461: this file is generated because it's a T4 template, so changes will be lost next time someone touches the .tt file. You should change .tt and re-generate.

greg0rym: This is done, however the change has changed the .cs file encoding to UTF-8-BOM (which matches the other cs files in the solution) and altered the whitespace. Is this a concern please? I believe I am using the latest version of TextTransform.exe for Visual Studio 2022. Thanks

aloneguid (approved):
LGTM, thank you!

aloneguid:
merging, will publish as a part of '-pre.8' nuget in a few minutes.

Issue #535, opened 2024-07-22
[BUG]: non null string fails with InvalidDataException and is wrongly flagged as nullable


ErikApption:
### Library Version

4.25.0-pre.2

### OS

Windows

### OS Architecture

64 bit

### How to reproduce?

1. Deserialize simple parquet file with schema (obtained from Tad - the parquet file also crashes floor)
![image](https://github.com/user-attachments/assets/1a4e5883-bf50-43f4-a2fd-f4fd3b332114)

2. "System.IO.InvalidDataException: 'property 'Value' is declared as 'value (System.String?)' but source data has it as 'value (System.String)''"

with 
```csharp
public class MyClass 
{
    [JsonPropertyName("id")]
    public long Id { get; set; }

    [JsonPropertyName("value")]
    public string Value { get; set; } = null!;

    [JsonPropertyName("frequency")]
    public double Frequency { get; set; }
}
```

The important part to note is that the exception is the same wherever Value is defined as nullable or not.
```csharp
    [JsonPropertyName("value")]
    public string? Value { get; set; } = null!;
``` 

ends up with exactly same error message. 

### Failing test

```csharp
var records = await ParquetSerializer.DeserializeAllAsync<MyClass>(stream).ToListAsync();
```

aloneguid:
I'm not sure how to reproduce this to be honest, are you trying to deserialise a particular file with class serializer? Do you have a sample?

ErikApption:
> I'm not sure how to reproduce this to be honest, are you trying to deserialise a particular file with class serializer? Do you have a sample?

I added unit tests in my PR to show the issue.

```csharp
        class PocoClassNotNullable {
            [JsonPropertyName("id")]
            public long Id { get; set; }

            [JsonPropertyName("value")]
            public string Value { get; set; } = null!;

            [JsonPropertyName("frequency")]
            public double Frequency { get; set; }
        }

        [Fact]
        public void Strings_NotNullable() {
            ParquetSchema s = typeof(PocoClassNotNullable).GetParquetSchema(true);
            Assert.False(s.DataFields[0].IsNullable);
            Assert.False(s.DataFields[1].IsNullable);
            Assert.False(s.DataFields[2].IsNullable);
        }
```

This fails with current version. The current code will consider any string as nullable, so I don't think you can deserialize a non nullable string in parquet to a class unless you add `[ParquetRequired]` to that field. This is because the Roslyn annotations are not used so any string is technically nullable.

[...]

PR #537, opened 2024-07-23
Support for Roslyn nullable annotations

ErikApption:
This is a fix for bug #535. The current parquet.net doesn't leverage compiler nullable when checking if members are nullable. This affects all classes - so any `string` property was marked as nullable in the schema. This PR adds checks for [Roslyn's annotations](https://github.com/dotnet/roslyn/blob/main/docs/features/nullable-metadata.md) to ensure that a field is nullable.

- Added checks for roslin compiler attributes to properly identify nullable and not nullable classes
- Added supporting unit tests

### DIFF
--- a/src/Parquet.Test/DataAnalysis/DataFrameReaderTest.cs
+++ b/src/Parquet.Test/DataAnalysis/DataFrameReaderTest.cs
@@ -12,19 +12,19 @@ namespace Parquet.Test.DataAnalysis {
     public class DataFrameReaderTest : TestBase {
 
         [Theory]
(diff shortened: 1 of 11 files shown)


aloneguid (approved):
LGTM, thanks!

aloneguid (approved):

ErikApption:
I could look at revising this pr to disable nullable support by default but any chances to consider changing the default in version 5? Working with files that are not generated by parquet.net is not intuitive without nullable support. There might be a compromise to limit issues and load null values into non nullable fields and simply show a warning.

Issue #542, opened 2024-08-21
[BUG]: parquet containing ConvertedType TIMESTAMP_MILLIS may throw IndexOutOfRangeException while reading.


artnim:
### Library Version

4.24.0

### OS

Ubuntu Linux

### OS Architecture

64 bit

### How to reproduce?

1. Create a parquet with timeseries as datetime64[ms]

```python
import pandas as pd

pd.set_option('io.parquet.engine', 'fastparquet')

start_time = pd.Timestamp('2024-08-21T00:00:00+02:00')
end_time = pd.Timestamp('2024-08-22T00:00:00+02:00')
series = pd.date_range(start_time, end_time, freq=pd.Timedelta(minutes=15))[:-1]

series = series.tz_convert('UTC').tz_localize(None).astype("datetime64[ms]")

df = pd.DataFrame({'datetime': series})

print(df)
print(df.dtypes)

df.to_parquet('./test.parquet')
```

2. Open the resulting test.parquet using floor
![image](https://github.com/user-attachments/assets/1464afb6-b335-48c4-95db-b861e5167047)

The same works like a charm reading a parquet with created with the library. So I suggest a dependency with the fastparquet engine used my site.

artnim:
Meanwhile, I can confirm that the problem is caused by what fastparquet writes. The problem does not exist with pyarrow.

PR #543, opened 2024-08-21
Fix return of Decode methods to returning the actual destination length.

artnim:
Solves #542 

Caused by `Decode` was returning the length of the source divided by the sizeof(T) instead of the length of the actually decoded data.


aloneguid (changes requested):
Looks good, but need to resolve conflicts.

artnim:
> Looks good, but need to resolve conflicts.

Just resolved the conflict around unneeded usings.

PR #521, opened 2024-06-16
feat: Logical Timestamp

cliedeman:
Resolves #492

https://github.com/apache/parquet-format/blob/master/LogicalTypes.md#timestamp

PR #546, opened 2024-09-09
feat: DateTimeOffset

cliedeman:
This is my follow up to #521 

Rationale: 
I have many models that currently use DateTimeOffset to map entities and I would like to serialize these without creating mapping models.

I am aware that DateTimeOffset was previously removed due to confusion but I have added additional checks to try and mitigate this.

The Logical Timestamp helps with this as we can refuse to decode a Timestamp to DateTimeOffset when it is not in UTC. As this invites errors and can cause confusion. (DateTime avoids this by allowing you to stipulate the kind as `Local`).

### DIFF
--- a/src/Parquet/Schema/DataField.cs
+++ b/src/Parquet/Schema/DataField.cs
@@ -153,13 +153,20 @@ namespace Parquet.Schema {
                 return false;
 
             return base.Equals(obj) &&
-                BaseClrType == other.BaseClrType &&
+                BaseClrTypeCompatible(other) &&
                 IsNullable == other.IsNullable &&
                 IsArray == other.IsArray;
         }
 
         /// <inheritdoc/>
         public override int GetHashCode() => base.GetHashCode();
+        
+        /// <summary>
+        /// 
+        /// </summary>
+        /// <param name="other"></param>
+        /// <returns></returns>
+        protected virtual bool BaseClrTypeCompatible(DataField other) => this.BaseClrType == other.BaseClrType;
 
         #region [ Type Resolution ]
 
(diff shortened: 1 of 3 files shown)


src/Parquet/Encodings/ParquetPlainEncoder.cs
```diff
@@ -900,6 +956,45 @@ public static void Encode(ReadOnlySpan<TimeOnly> data, Stream destination, Schem
...
         }
 #endif
+        public static int Decode(Span<byte> source, Span<DateTimeOffset> data, SchemaElement tse) {
+            switch(tse.Type) {
+                case TType.INT64:
+                    long[] longs = ArrayPool<long>.Shared.Rent(data.Length);
+                    try {
+                        int longsRead = Decode(source, longs.AsSpan(0, data.Length));
+                        if(tse.LogicalType?.TIMESTAMP is not null) {
+                            if(!tse.LogicalType.TIMESTAMP.IsAdjustedToUTC) {
```
cliedeman, line 966: This probably deserved a dedicated error and a link to the docs if this change is accepted

src/Parquet/Schema/DataField.cs
```diff
@@ -158,13 +153,20 @@ public override bool Equals(object? obj) {
...
 
         /// <inheritdoc/>
         public override int GetHashCode() => base.GetHashCode();
+        
+        /// <summary>
+        /// 
+        /// </summary>
+        /// <param name="other"></param>
+        /// <returns></returns>
+        protected virtual bool BaseClrTypeCompatible(DataField other) => this.BaseClrType == other.BaseClrType;
```
cliedeman, line 169: This is a hack to fix the serialization append issue I encountered.

Fundamentally the parquet thrift schema on its own does not enough information to differentiate date time offset vs date time with the class.

Will have to rethink my approach. Possibly merge DateTimeOffsetDataField into DateTimeDataField somehow

PR #547, opened 2024-09-09
fix: Make isAdjustedToUTC defualt to true and allow it to be set

cliedeman:
Missed `IsAdjustedToUTC` in `ParquetTimestampAttribute`.
Default it to true because a local timestamp should be the exception.

Haven't thought through exactly how this would behave backwards compat. Will try to find time to setup a test where we convert from a non logical timestamp to a logical one to check the behaviour


aloneguid (approved):
LGTM, thanks!

PR #551, opened 2024-09-12
fix: Handle nullable enums

cliedeman:
Nullable Enums are not correctly unwrapped to the primitive type


aloneguid (approved):
LGTM, thanks!

PR #566, opened 2024-10-18
Added untyped async enumable deserialization

flambert860:
Added untyped async enumable deserialization so that we do not have to load the whole data into memory.


aloneguid (approved):
This looks good to me, thank you!

PR #597, opened 2025-02-20
[#436] Expose `RecyclableMemoryStreamManager` memory settings

### LINKED ISSUE #436: Possible memory leak in DataColumnWriter
While investigating a potential memory leak in my Azure Function app, I noticed that the Managed Memory Tool in Visual Studio tells me that I have many objects in the .NET managed heap. When looking through those objects (`byte[]`), they seem to have `Microsoft.IO.RecyclableMemoryStreamManager` as their root.

Parquet.Net uses `RecyclableMemoryStreamManager` in [`DataColumnWriter`](https://github.com/aloneguid/parquet-dotnet/blob/9dcd02ec3e2120d31dd85ed31afc9255c2fc8f5c/src/Parquet/File/DataColumnWriter.cs#L24). 

In it's [documentation](https://github.com/microsoft/Microsoft.IO.RecyclableMemoryStream#usage-guidelines) it states that

> Important!: If you do not set MaximumFreeLargePoolBytes and MaximumFreeSmallPoolBytes there is the possibility for unbounded memory growth!

And [also](https://github.com/microsoft/Microsoft.IO.RecyclableMemoryStream#how-it-works) that:

>  the RecyclableMemoryStreamManager will use the properties MaximumFreeSmallPoolBytes and MaximumFreeLargePoolBytes to determine whether to put those buffers back in the pool, or let them go (and thus be garbage collected). It is through these properties that you determine how large your pool can grow. If you set these to 0, you can have unbounded pool growth, which is essentially indistinguishable from a memory leak.

It seems that the `DataColumnWriter` does not set these properties, so I guess that might be the reason for my app's high memory usage.

Should those `MaximumFreeSmallPoolBytes` and `MaximumFreeLargePoolBytes` properties be somehow user configurable? Maybe via [`ParquetOptions`](https://github.com/aloneguid/parquet-dotnet/blob/9dcd02ec3e2120d31dd85ed31afc9255c2fc8f5c/src/Parquet/ParquetOptions.cs#L8)?

DrewMcArthur:
Adds two fields to `ParquetOptions`: `MaximumSmallPoolFreeBytes` and `MaximumLargePoolFreeBytes`, which correspond to and update the settings of the same name in the `DataColumnWriter`'s `RecyclableMemoryStreamManager _rmsMgr` instance.

Will leave a handful of open questions as conversations on the code.

closes #436

### DIFF
--- a/src/Parquet/File/DataColumnWriter.cs
+++ b/src/Parquet/File/DataColumnWriter.cs
@@ -38,6 +38,8 @@ namespace Parquet.File {
             _compressionLevel = compressionLevel;
             _keyValueMetadata = keyValueMetadata;
             _options = options;
+            _rmsMgr.Settings.MaximumSmallPoolFreeBytes = options.MaximumSmallPoolFreeBytes;
+            _rmsMgr.Settings.MaximumLargePoolFreeBytes = options.MaximumLargePoolFreeBytes;
         }
 
         public async Task<ColumnChunk> WriteAsync(
--- a/src/Parquet/ParquetOptions.cs
+++ b/src/Parquet/ParquetOptions.cs
@@ -57,5 +57,39 @@ namespace Parquet {
         /// your readers do not understand it.
         /// </summary>
         public bool UseDeltaBinaryPackedEncoding { get; set; } = true;
+
+        /// <summary>
+        /// This option is passed to the <see cref="Microsoft.IO.RecyclableMemoryStreamManager"/> , 
+        /// which keeps a pool of streams in memory for reuse. 
+        /// By default when this option is unset, the RecyclableStreamManager 
+        /// will keep an unbounded amount of memory, which is 
+        /// "indistinguishable from a memory leak" per their documentation.
+        /// 
+        /// This does not restrict the size of the pool, but just allows 
+        /// the garbage collector to free unused memory over this limit.
+        /// 
+        /// You may want to adjust this smaller to reduce max memory usage, 
+        /// or larger to reduce garbage collection frequency.
+        /// 
+        /// Defaults to 16MB.  
+        /// </summary>
+        public int MaximumSmallPoolFreeBytes { get; set; } = 16 * 1024 * 1024;
+
+        /// <summary>
+        /// This option is passed to the <see cref="Microsoft.IO.RecyclableMemoryStreamManager"/> , 
+        /// which keeps a pool of streams in memory for reuse. 
+        /// By default when this option is unset, the RecyclableStreamManager 
+        /// will keep an unbounded amount of memory, which is 
+        /// "indistinguishable from a memory leak" per their documentation.
+        /// 
+        /// This does not restrict the size of the pool, but just allows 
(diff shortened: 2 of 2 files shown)


src/Parquet/File/DataColumnWriter.cs
```diff
@@ -38,6 +38,8 @@ public DataColumnWriter(
             _compressionLevel = compressionLevel;
             _keyValueMetadata = keyValueMetadata;
             _options = options;
+            _rmsMgr.Settings.MaximumSmallPoolFreeBytes = options.MaximumSmallPoolFreeBytes;
+            _rmsMgr.Settings.MaximumLargePoolFreeBytes = options.MaximumLargePoolFreeBytes;
```
DrewMcArthur, line 42: the `_rmsMgr` is static, but the `options` are not - so this leads to two things.

1. these settings are not configured when the instance is created.  is that a problem?
2. each time we instantiate a `DataColumnWriter`, we'll needlessly run these two lines to update the settings, when they really only need to be run once.  I assume the overhead shouldn't matter too much, unless we're creating a lot of these writers inside of a hot loop.

src/Parquet/ParquetOptions.cs
```diff
@@ -57,5 +57,39 @@ public class ParquetOptions {
...
+        /// 
+        /// This does not restrict the size of the pool, but just allows 
+        /// the garbage collector to free unused memory over this limit.
+        /// 
+        /// You may want to adjust this smaller to reduce max memory usage, 
+        /// or larger to reduce garbage collection frequency.
+        /// 
+        /// Defaults to 64MB.
+        /// </summary>
+        public int MaximumLargePoolFreeBytes { get; set; } = 64 * 1024 * 1024;
```
DrewMcArthur, line 93: How do we feel about these names & defaults?

Part of me wants to just expose a `RecyclableMemoryStreamOptions` field that would give the user full control, but maybe that's overkill.  Another option is to put these two fields into their own struct, e.g. 

```c#
public class RecyclableStreamManagerOptions { // or alternatively, just `MemoryOptions`
    public int MaximumSmallPoolFreeBytes { get; set; } = 16 * 1024 * 1024;
    public int MaximumLargePoolFreeBytes { get; set; } = 64 * 1024 * 1024;
}
...
public class ParquetOptions {
    ...
    public RecyclableStreamManagerOptions RecyclableStreamManagerOptions { get; set; } = new();
}
```

doing something like this would reduce the duplicated documentation.  let me know how you feel about how i wrote that up, if you'd like a link back to #436 for context, etc.  also feel free to edit as you see fit!

DrewMcArthur (commented):
Some open questions for @aloneguid

src/Parquet/ParquetOptions.cs
aloneguid: I think this looks good, can't see any issues with it. I will merge into next release.

aloneguid (approved):
LGTM

PR #598, opened 2025-02-20
Fix `ParquetSchema` equality

### LINKED ISSUE #596: [BUG]: `ParquetSchema` equality collisions
### Library Version

5.1.0

### OS

Linux

### OS Architecture

64 bit

### How to reproduce?

1. Use Parquet.Net in a persistent app/service. The persistent part is important because of the caching in `GetAssembler` [here](https://github.com/aloneguid/parquet-dotnet/blob/master/src/Parquet/Serialization/ParquetSerializer.cs#L478-L481).
2. For the simplest example, prepare 2 parquet files where both files:
    - Have one column of the same name (ex: `example_column`)
    - Populate that column with a `Struct`, **but use different schemas within that `Struct`**
3. Use `DeserializeAllAsync` (untyped) to read one file and then the other
4. Get some error that looks roughly like:

```
System.InvalidOperationException: failed to deserialize column 'example_column/sample_field_1', pseudo code: ['.Block() { .Loop { .Block() { .If ( $dlIdx < ($dc.DefinitionLevels).Length | $dataIdx < ($dc.DefinedData).Length ) { .Block() { $dl = ($dc.DefinitionLevels)[$dlIdx++]; .If ($dl == 2) { $dataElement = $data[$dataIdx++] } .Else { .Default(System.Void) }; .Default(System.Void); $hasData = True } } .Else { $hasData = False }; .If ( .IsFalse($hasData) ) { .Break #Label1 { } } .Else { .Default(System.Void) }; .Default(System.Void); .Block() { .If ($dl >= 1) { .Block() { .If ( !.Call $curr.ContainsKey("example_column") ) { $curr.Item["example_column"] = .New System.Collections.Generic.Dictionary`2[System.String,System.Object]() } .Else { .Default(System.Void) }; .Block(System.Collections.Generic.Dictionary`2[System.String,System.Object] $var1) { $var1 = (System.Collections.Generic.Dictionary`2[System.String,System.Object])$curr.Item["example_column"]; .Block() { .If (2 == $dl) { $var1.Item["sample_field_1"] = (System.Object)$dataElement } .Else { .Default(System.Void) } } } } } .Else { .Default(System.Void) } }; .If (0 == 0) { .Break #Label1 { } } .Else { .Default(System.Void) } } } .LabelTarget #Label1: }'] 
---> System.IndexOutOfRangeException: Index was outside the bounds of the array. 
at void lambda_method10548(Closure, IEnumerable<Dictionary<string, object>>, DataColumn) 
at async Task Parquet.Serialization.ParquetSerializer.DeserializeRowGroupAsync<T>(ParquetRowGroupReader rg, ParquetSchema schema, Assembler<T> asm, ICollection<T> result, ParquetSerializerOptions options, CancellationToken cancellationToken, bool resultsAlreadyAllocated) --- End of inner exception stack trace --- 
at async Task Parquet.Serialization.ParquetSerializer.DeserializeRowGroupAsync<T>(ParquetRowGroupReader rg, ParquetSchema schema, Assembler<T> asm, ICollection<T> result, ParquetSerializerOptions options, CancellationToken cancellationToken, bool resultsAlreadyAllocated) 
at async Task Parquet.Serialization.ParquetSerializer.DeserializeRowGroupAsync(ParquetReader reader, int rgi, Assembler<Dictionary<string, object>> asm, List<Dictionary<string, object>> result, ParquetSerializerOptions options, CancellationToken cancellationToken) 
at async IAsyncEnumerable<Dictionary<string, object>> Parquet.Serialization.ParquetSerializer+<>c__DisplayClass20_0.<DeserializeAllAsync>g__GetAsyncRowGroups|0(?)+GetAsyncRowGroups() 
at async Task<(List<ParquetColumn> Columns, List<List<string>> Rows)>
```

I've debugged this locally and it appears that the issue is due to [how IEquatable](https://github.com/aloneguid/parquet-dotnet/blob/master/src/Parquet/Schema/ParquetSchema.cs#L225-L251) is implemented for `ParquetSchema`. [Equals](https://github.com/aloneguid/parquet-dotnet/blob/master/src/Parquet/Schema/ParquetSchema.cs#L153-L159) and [GetHashCode](https://github.com/aloneguid/parquet-dotnet/blob/master/src/Parquet/Schema/ParquetSchema.cs#L249-L251) are both based on the `_fields`, but [the implementation for IEquatable for Fields](https://github.com/aloneguid/parquet-dotnet/blob/master/src/Parquet/Schema/Field.cs#L139-L149) just cares about `Name` and `Path` (which will be identical for `Fields`) and `SchemaType`, which will both be `SchemaType.Struct` in the example above.

It seems like this would be easily fixable by using the `DataFields` instead since that will include [more robust equality checks](https://github.com/aloneguid/parquet-dotnet/blob/master/src/Parquet/Schema/DataField.cs#L154-L162) and include deeper `Path` checks.

### Failing test

```csharp
[Fact]
public void Schemas_different_structs_not_equal() {
    var schema1 = new ParquetSchema(new StructField("struct", new DataField<int>("id"), new DataField<string>("city")));
    var schema2 = new ParquetSchema(new StructField("struct", new DataField<int>("id"), new DataField<string>("city2")));

    Assert.NotEqual(schema1, schema2);
}
```

goeffthomas:
Fixes #596

Also included a drive-by fix to the contributing guide, which was missing the owner name in the repo URL

### DIFF
--- a/.github/CONTRIBUTING.md
+++ b/.github/CONTRIBUTING.md
@@ -31,7 +31,7 @@ At this point, you should switch back to your master branch and make sure it's
 up to date with *Parquet.Net* master branch:
 
 ```sh
-git remote add upstream git@github.com:parquet-dotnet.git
+git remote add upstream git@github.com:aloneguid/parquet-dotnet.git
 git checkout master
 git pull upstream master
 ```
--- a/src/Parquet.Test/Schema/SchemaTest.cs
+++ b/src/Parquet.Test/Schema/SchemaTest.cs
@@ -8,8 +8,6 @@ using Parquet.Schema;
 using System.Linq;
 using TT = Parquet.Meta.Type;
 using CT = Parquet.Meta.ConvertedType;
-using System.Numerics;
-using Parquet.Encodings;
 using Parquet.File;
 using Parquet.Meta;
 
@@ -36,7 +34,7 @@ namespace Parquet.Test.Schema {
         }
 
         [Fact]
-        public void Schemas_idential_equal() {
+        public void Schemas_identical_equal() {
             var schema1 = new ParquetSchema(new DataField<int>("id"), new DataField<string>("city"));
             var schema2 = new ParquetSchema(new DataField<int>("id"), new DataField<string>("city"));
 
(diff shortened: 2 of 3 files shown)


aloneguid (approved):
LGTM, thanks!

PR #602, opened 2025-03-04
Handle decimals with no scale defined: DECIMAL(X, 0)

mukunku:
# Summary

One of my users pointed out that parquet files created with Xtract Universal, a SAP plugin, can't be opened if it contains Decimal values with a scale of zero: https://github.com/mukunku/ParquetViewer/issues/128

This is because while the plugin is setting the Precision of the decimal it's not setting the Scale, leaving it null. 

This PR suggests changes so we can handle such decimal schema definitions gracefully.


src/Parquet/Encodings/SchemaEncoder.cs
```diff
@@ -265,7 +265,7 @@ private static bool TryBuildDataField(SchemaElement se, ParquetOptions options,
         private static DataField GetDecimalDataField(SchemaElement se) =>
             new DecimalDataField(se.Name,
                 se.Precision.GetValueOrDefault(DecimalFormatDefaults.DefaultPrecision),
-                se.Scale.GetValueOrDefault(DecimalFormatDefaults.DefaultScale));
+                se.Scale.GetValueOrDefault(se.Precision is not null ? 0 : DecimalFormatDefaults.DefaultScale));
```
mukunku, line 268: In my test file the Precision is 9 and the Scale is 0. With `DefaultScale` here it tries to construct a decimal like DECIMAL(9,18) which crashes since the scale can't be larger than the precision. In my opinion if there's a precision defined, but not a scale, assuming it's 0 makes more sense.

src/Parquet/File/Values/Primitives/BigDecimal.cs
```diff
@@ -30,7 +30,7 @@ public BigDecimal(byte[] data, SchemaElement schema, bool isReversed = false) {
 
             UnscaledValue = new BigInteger(data);
             Precision = schema.Precision!.Value;
-            Scale = schema.Scale!.Value;
+            Scale = schema.Scale ?? 0;
```
mukunku, line 33: Similar to the logic above: if there's a precision, don't crash if there's no scale and assume it's 0.

aloneguid (approved):
looks goot to me, sorry for the delay

PR #616, opened 2025-05-24
Fix parsing empty byte array.

Kuinox:
Currently when a row contain an empty byte array, parquet.net throw an IndexOutOfRangeException.  
This PR add a test for this case and fix the bug.


Kuinox:
@aloneguid the 5.1.2-pre.1 do not contain the change of this PR:  

image

PR #631, opened 2025-07-20
Fix #630 Calculate empty and null DL for nullable items

### LINKED ISSUE #630: [BUG]: Serialize doesn't handle empty + null list of nullable type
### Library Version

5.1.1

### OS

All

### OS Architecture

ARM 64

### How to reproduce?

When serializing a list of strings (or other nullable item type), the parquet serializer will serialize:
* a null list to an empty list
* an empty list to a list with one null item.

This is because FieldStriperCompiler uses [itemMaxDL - 2](https://github.com/aloneguid/parquet-dotnet/blob/master/src/Parquet/Serialization/Dremel/FieldStriperCompiler.cs#L323) and [itemMaxDL - 1](https://github.com/aloneguid/parquet-dotnet/blob/master/src/Parquet/Serialization/Dremel/FieldStriperCompiler.cs#L329) for null and empty list. However, this only works when the item type is non-nullable.

Suggested fix: use `listMaxDL - 2` and `listMaxDL - 1`. This should work regardless of item nullability.

### Failing test

```csharp
public class TestRow
{
    public int ID { get; set; }
    public List<int> Numbers { get; set; } = []; // This serializes properly
    public List<string> Words { get; set; } = []; // This doesn't
}

public class Program
{
    public static async Task Main(string[] args)
    {
        TestRow[] rows = [
            new TestRow { ID = 1, Numbers = [456], Words = ["abc"]},
            new TestRow { ID = 2, Numbers = [], Words = []},
            new TestRow { ID = 3, Numbers = [null], Words = [null]},
            new TestRow { ID = 4, Numbers = null, Words = null}
        ];
        string testFile = "test.parquet";

        await ParquetSerializer.SerializeAsync(rows, testFile);
    }

}



/*
Pyspark code:
df = spark.read.format("parquet").load("test.parquet")
df.show()

*/
```

mrinal-thomas:
Fixes #630 .

Note, I think this case also doesn't work properly for deserialization, but I didn't look into that much (not needed for my use case).

### DIFF
--- a/src/Parquet.Test/Serialisation/AdditionalDremelStriperTests.cs
+++ b/src/Parquet.Test/Serialisation/AdditionalDremelStriperTests.cs
@@ -1,0 +1,48 @@
+using System.Collections.Generic;
+using Parquet.Serialization;
+using Parquet.Serialization.Dremel;
+using Parquet.Test.Serialisation.Paper;
+using Xunit;
+
+namespace Parquet.Test.Serialisation {
+    public class TestRow {
+        public List<int>? Numbers { get; set; } = [];
+        public List<string?>? Words { get; set; } = [];   
+    }
+
+    /// <summary>
+    /// Additional tests beyond the example given in the google paper, but needed for completeness.
+    /// </summary>
+    public class AdditionalDremelStriperTests {
+        private readonly Striper<TestRow> _striper;
+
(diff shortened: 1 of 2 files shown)


aloneguid (approved):
LGTM, thanks

PR #652, opened 2025-09-22
Add implementation for bloom filters (predicate pushdown)

kiloOhm:
As described in the [specification](https://parquet.apache.org/docs/file-format/bloomfilter/)

With the changes in this PR you can now add column-specific settings for bloom filters:

``` c#
new ParquetOptions {
    BloomFilterOptionsByColumn = new Dictionary<string, ParquetOptions.BloomFilterOptions>() {
        { field.Name, new ParquetOptions.BloomFilterOptions { 
            EnableBloomFilters = true,
            BloomFilterFpp = 0.01f,
        } }
    }
},
```

During writing, the bloom filters are constructed and written to the data column header

While reading, you can then check if a value is definitely not or maybe present in the rowgroup column:

``` c#
var reader = new DataColumnReader(field, ms, chunk, stats, footer, parquetOptions: new ParquetOptions {
    BloomFilterOptionsByColumn = new Dictionary<string, ParquetOptions.BloomFilterOptions>() {
            { field.Name, new ParquetOptions.BloomFilterOptions { EnableBloomFilters = true } }
        }
});

Assert.True(reader.MightMatchEquals("contained"));
```


kiloOhm:
Just read the LLM rule. Closing

aloneguid:
@kiloOhm just to clarify it, I don't care or know what tools you use locally. The policy applies only to fully LLM-driven contributions where authors don't even understand what they are doing. Generating code-shaped text is not what I'm looking for.

PR #480, opened 2024-02-25
[WIP] Modular Encryption Support When Reading Parquet Files

mukunku:
# Summary

I made significant progress on getting [Footer Decryption](https://github.com/apache/parquet-format/blob/master/Encryption.md#54-encrypted-footer-mode) working with parquet files (#191).

I'm opening this work-in-progress pull request with hopes that some other folks can help get this across the finish line. 

## AES_GCM_V1

Thanks to a test file file @pzatschl [shared with me](https://github.com/mukunku/ParquetViewer/discussions/92) I was able to implement the [Aes Gcm V1 encryption algorithm](https://github.com/apache/parquet-format/blob/master/Encryption.md#421-aes_gcm_v1). 

[link to code](https://github.com/mukunku/parquet-dotnet-DecryptFile_UTF8_AesGcmV1_192bit/blob/209c60f8cb7f3f8aa83cb9aef3da6f957d7fd5cd/src/Parquet/Encryption/AES_GCM_V1.cs#L32-L60)

## AES_GCM_CTR_V1

I also implemented the [Aes Gcm Ctr V1 encryption algorithm](https://github.com/apache/parquet-format/blob/master/Encryption.md#422-aes_gcm_ctr_v1), however I don't have any test files to confirm it's working 🙃 

[link to code](https://github.com/mukunku/parquet-dotnet-DecryptFile_UTF8_AesGcmV1_192bit/blob/209c60f8cb7f3f8aa83cb9aef3da6f957d7fd5cd/src/Parquet/Encryption/AES_GCM_CTR_V1.cs#L28-L71)

# How to test

Checkout the unit test I added that tests the sample file I mentioned above:
[link to code](https://github.com/mukunku/parquet-dotnet-DecryptFile_UTF8_AesGcmV1_192bit/blob/873d528d2d0d1cf5e55978ed4491eb0bf077a072/src/Parquet.Test/ParquetReaderOnTestFilesTest.cs#L177-L192)
image

However, even though I can decrypt the test file successfully, the data itself doesn't seem to be valid. So I had to add this try-catch as a temporary workaround. 
[link to code](https://github.com/mukunku/parquet-dotnet-DecryptFile_UTF8_AesGcmV1_192bit/blob/873d528d2d0d1cf5e55978ed4491eb0bf077a072/src/Parquet/Extensions/OtherExtensions.cs#L16-L23)
We should remove this once we have a proper test file. (Unfortunately I don't have any other test files )

PR #654, opened 2025-09-27
Modular Encryption

kiloOhm:
This PR adds Modular Encryption as specified here: https://github.com/apache/parquet-format/blob/master/Encryption.md

It only covers the modules that have been implemented so far. So no Indexes, Offsets or Bloom Filters. Though these should be trivial to add later.

#480 was my starting point but there is only some file structure and architecture left of the original PR. I'm opening a separate PR so that @mukunku doesn't have to go through all my changes and merge before review and collaboration on this PR can start. 

@Mukunku 's work was very helpful though, so big thanks!

Functionality should be apparent from this Unit Test:

``` c#
[Fact]
public async Task B_PageEncryption_CtrVariant_RoundTrip() {
    ParquetSchema schema = MakeSchema();
    var opts = new ParquetOptions {
        SecretKey = "sixteen-byte-key",
        AADPrefix = "suite-b",
        SupplyAadPrefix = false,
        UseCtrVariant = true
    };

    using var ms = new MemoryStream();
    using(ParquetWriter writer = await ParquetWriter.CreateAsync(schema, ms, formatOptions: opts)) {
        using ParquetRowGroupWriter rg = writer.CreateRowGroup();
        var sField = (DataField)schema.Fields[0];
        await rg.WriteColumnAsync(new DataColumn(sField, Enumerable.Range(0, 100).Select(i => i.ToString()).ToArray()));
    }

    ms.Position = 0;
    using ParquetReader reader = await ParquetReader.CreateAsync(ms, new ParquetOptions {
        SecretKey = opts.SecretKey
    });

    using ParquetRowGroupReader rgr = reader.OpenRowGroupReader(0);
    var rf = (DataField)reader.Schema.Fields[0];
    DataColumn col = await rgr.ReadColumnAsync(rf);
    Assert.Equal(100, col.Data.Length);
    Assert.Equal("0", (string)col.Data.GetValue(0)!);
    Assert.Equal("99", (string)col.Data.GetValue(99)!);
}
```

I wrote a little java tool to verify compatibility with parquet-mr. I added it to. the tools folder: tools/encrypted-parquet-inspector-1.0.0.jar

Source can be found here: https://github.com/kiloOhm/parquet-mr-tools/blob/main/encrypted-parquet-inspector/src/main/java/org/example/ParquetInspect.java

There are a hand full of integration tests using this tool along with a bunch of Unit Tests.
Further integration tests would be appreciated.


kiloOhm:
Just realized that there is a bunch of stuff missing from the spec. Will reopen when it's actually complete

aloneguid:
Thanks, I'll pick this up soon.

aloneguid:
Closing as it has drifted from master quite a lot and probably will be easier to start fresh based on ideas here.

PR #656, opened 2025-10-01
Add validation to prevent writing malformed Parquet files

Kuinox:
Disclaimer: this commit was generated with an AI (not this PR description), but I find it good enough.  

When upgrading my own tool to add 2 new columns i forgot twice something which led to generating bad parquets file (which also tripped an assert in DuckDB).

There is a lack of validation of two things, that's allowed me to write a parquet with 
- 2 columns missing.
- 2 columns with a different row count than the rest of the row group.

This commit check theses problem and throw to not produced invalid parquets.


aloneguid (approved):
LGTM, thank you

PR #685, opened 2025-12-02
Add write benchmark of NYC Taxi Dataset.

Kuinox:
Since I want to do work to improve the performance of this library, I figured benching against real world data instead of synthetic data was needed.  
This would avoid me writing optimisations for my specific use case and ruin performance for other scenarios.  

```
BenchmarkDotNet v0.15.7, Windows 11 (10.0.26100.7171/24H2/2024Update/HudsonValley)
AMD Ryzen 9 7900X 4.70GHz, 1 CPU, 24 logical and 12 physical cores
.NET SDK 10.0.100
  [Host]   : .NET 8.0.22 (8.0.22, 8.0.2225.52707), X64 RyuJIT x86-64-v4
  ShortRun : .NET 8.0.22 (8.0.22, 8.0.2225.52707), X64 RyuJIT x86-64-v4

Job=ShortRun  IterationCount=3  LaunchCount=1
WarmupCount=3

| Method                         | Dataset        | Mean       | Error    | StdDev   | Allocated  |
|------------------------------- |--------------- |-----------:|---------:|---------:|-----------:|
| 'Parquet.Net -> MemoryStream'  | tripdata       |   466.8 ms | 383.5 ms | 21.02 ms |  727.13 MB |
| 'ParquetSharp -> MemoryStream' | tripdata       |   256.4 ms | 446.9 ms | 24.50 ms |   44.78 MB |
| 'Parquet.Net -> Disk'          | tripdata       |   453.2 ms | 369.3 ms | 20.24 ms |  697.19 MB |
| 'ParquetSharp -> Disk'         | tripdata       |   275.8 ms | 341.0 ms | 18.69 ms |   11.61 MB |
| 'Parquet.Net -> MemoryStream'  | tripdata-large | 1,144.2 ms | 981.2 ms | 53.78 ms | 2103.57 MB |
| 'ParquetSharp -> MemoryStream' | tripdata-large |   585.0 ms | 510.3 ms | 27.97 ms |  172.53 MB |
| 'Parquet.Net -> Disk'          | tripdata-large | 1,166.8 ms | 923.9 ms | 50.64 ms |  2007.4 MB |
| 'ParquetSharp -> Disk'         | tripdata-large |   591.0 ms | 213.6 ms | 11.71 ms |   38.24 MB |
```

Sadly the benchmark shows Parquet.Net is up to 2x slower than parquet sharp.  

I see two way of improving the performance:  
- Parallelisation  
- Improving the encoding algos  

## Parallelisation  
I believe we can run mosts of the compute intensive stuff in parallel. The users would then just do Task.WhenAll.

## Improving the encoding algos

When I profiled for my use case, most of the time was spent on encoding, and a significant portion on deduplicating the input.  
I believe there is room for improvement here and easily earn 10-20%.


Kuinox:
Did both ran with the same runtime ?

aloneguid:
Yeah we know there is a lot of room for improvement. Especially since delta encoding was enabled by default, which is CPU heavy on write. This would be my focus for the near future.

The issue with these benchmarks is reliability. We don't know which compression is used, what encodings are utilised by each library, what's the file size, how efficient the target file is and so on. This is just testing defaults, which I can change for existing parquet.net and get at least 10x speed increase. It will be more useful to have benchmarks which test two libraries in exactly the same conditions. It's hard to achieve though.

Kuinox:
Good point, i'll set the same compression algo for each bench

aloneguid:
I don't know how it's going to work though without re-testing each commit at once. Action agents allocate random node with unpredictable performance.

aloneguid:
V6 preview (in master) had a lot of improvements, I'd recommend looking into it.

Kuinox:
@aloneguid On another project I have some benchmark and I see 50% perf regression for plain encoding (except bool).
Also it looks like I can't force Parquet.Net on Dictionnary encoding on V6 preview.  

btw the other project in question is a .NET Parquet lib :p

aloneguid:
@Kuinox that's interesting, because all the tests I came up so far show massive performance win v5 vs v6. I'd be interested to see how you benchmark.
P.S. You can force dictionary encoding per column in `ParquetOptions`.

Kuinox:
this doesn't produce a parquet that use dictionary encoding:
```csharp
#:package Parquet.Net@6.0.0-pre.8

using Parquet;
using Parquet.Schema;

var field = new DataField<int>("value");
var schema = new Parquet.Schema.ParquetSchema(field);
var options = new ParquetOptions
{
    CompressionMethod = CompressionMethod.None,
    DictionaryEncodingThreshold = 1.0,
    DictionaryEncodingSampleSize = 0
};
options.ColumnEncodingHints.Add(field.Path.ToString(), EncodingHint.Dictionary);

var values = Enumerable.Range(0, 4_096).Select(static i => i % 16).ToArray();

await using var stream = File.Create("parquetnet-int32-dictionary-hint.parquet");
await using var writer = await Parquet.ParquetWriter.CreateAsync(schema, stream, options, false);
using var rowGroup = writer.CreateRowGroup();
await rowGroup.WriteAsync<int>(field, values.AsMemory(), null, null, default);
```

Kuinox:
Huh, I asked codex to update & run benchs and it decided to puch the branch by itself 🤡

Kuinox:
The self comparison bench added by codex in this commit have the same result that what I found:  


| Method               | Job                 | Arguments               | Dataset        | LogicalEncoding   | Mean       | Error      | StdDev    | Gen0        | Gen1      | Gen2      | Allocated |
|--------------------- |-------------------- |------------------------ |--------------- |------------------ |-----------:|-----------:|----------:|------------:|----------:|----------:|----------:|
| **&#39;Parquet.Net source&#39;** | **Parquet.Net (local)** | **Default**                 | **tripdata**       | **Plain**             | **1,039.3 ms** |   **412.2 ms** |  **22.59 ms** |  **33000.0000** |         **-** |         **-** |   **1.76 GB** |
| &#39;Parquet.Net source&#39; | Parquet.Net 5.6.0   | /p:ParquetVersion=5.6.0 | tripdata       | Plain             |   891.5 ms |   455.5 ms |  24.97 ms |  33000.0000 |         - |         - |   1.76 GB |
| **&#39;Parquet.Net source&#39;** | **Parquet.Net (local)** | **Default**                 | **tripdata**       | **RleDictionary**     | **1,006.4 ms** |   **390.5 ms** |  **21.40 ms** |  **22000.0000** |         **-** |         **-** |   **1.77 GB** |
| &#39;Parquet.Net source&#39; | Parquet.Net 5.6.0   | /p:ParquetVersion=5.6.0 | tripdata       | RleDictionary     | 1,793.1 ms |   370.4 ms |  20.31 ms |  12000.0000 |         - |         - |   2.39 GB |
| **&#39;Parquet.Net source&#39;** | **Parquet.Net (local)** | **Default**                 | **tripdata**       | **DeltaBinaryPacked** |   **927.2 ms** |   **569.7 ms** |  **31.23 ms** |  **35000.0000** |         **-** |         **-** |   **1.07 GB** |
| &#39;Parquet.Net source&#39; | Parquet.Net 5.6.0   | /p:ParquetVersion=5.6.0 | tripdata       | DeltaBinaryPacked |   768.4 ms |   416.3 ms |  22.82 ms |  35000.0000 |         - |         - |   1.07 GB |
| **&#39;Parquet.Net source&#39;** | **Parquet.Net (local)** | **Default**                 | **tripdata-large** | **Plain**             | **3,724.5 ms** | **3,682.9 ms** | **201.87 ms** | **108000.0000** | **2000.0000** | **1000.0000** |   **7.23 GB** |
| &#39;Parquet.Net source&#39; | Parquet.Net 5.6.0   | /p:ParquetVersion=5.6.0 | tripdata-large | Plain             | 3,208.4 ms | 1,615.8 ms |  88.57 ms | 107000.0000 | 1000.0000 |         - |   6.17 GB |
| **&#39;Parquet.Net source&#39;** | **Parquet.Net (local)** | **Default**                 | **tripdata-large** | **RleDictionary**     | **3,779.6 ms** |   **451.4 ms** |  **24.74 ms** |  **73000.0000** |         **-** |         **-** |   **7.24 GB** |
| &#39;Parquet.Net source&#39; | Parquet.Net 5.6.0   | /p:ParquetVersion=5.6.0 | tripdata-large | RleDictionary     | 6,035.2 ms | 5,437.7 ms | 298.06 ms |  37000.0000 |         - |         - |   7.63 GB |
| **&#39;Parquet.Net source&#39;** | **Parquet.Net (local)** | **Default**                 | **tripdata-large** | **DeltaBinaryPacked** | **3,001.3 ms** | **3,546.4 ms** | **194.39 ms** | **114000.0000** | **1000.0000** |         **-** |   **3.87 GB** |
| &#39;Parquet.Net source&#39; | Parquet.Net 5.6.0   | /p:ParquetVersion=5.6.0 | tripdata-large | DeltaBinaryPacked | 2,686.9 ms | 1,473.0 ms |  80.74 ms | 114000.0000 | 1000.0000 |         - |   4.08 GB |

aloneguid:
API in v6 is not compatible with v5, so I'm not quite sure how LLM made those changes work.

Kuinox:
It sprinkled with #if a bit everywhere, and it very ugly.  

I can make it better, but first, do you even want to support benching against the previous version ?

aloneguid:
Not really, only wondering why your tests are slower on v6. My approach was to run benchmarks on v5 branch, then on master. Separately one after another on the same machine.

Kuinox:
Ok, do you want I keep digging into a more minimal benchmark/dataset to compare the two ?

aloneguid:
Not at all, I was just wondering how 5 is slower, I couldn't reproduce this with a test.

PR #695, opened 2025-12-27
Fix DateTimeKind collisions in dictionary encoding and mark INT96 DateTimeKind as Unspecified

Kuinox:
While I'm trying to fix [the DateTime handling](https://github.com/aloneguid/parquet-dotnet/pull/685#issuecomment-3679111320), I'm uncovering other bugs.  

The one fixed in this PR, is when there are mixed DateTimeKind in a RowGroup.
The DateTime are merged by tick count without taking into account the DateTimeKind.  
When the DateTimeKind is local, it represent a local date and while having the same TickCount, it represent another moment of time than an utc DateTime with the same TickCount.

The root cause of the issue, is that .NET DateTime are equal if they have the same TickCount, but different kind.

This PR also change the DateTimeKind we put on decoded DateTime from INT96. They are now Unspecified, instead of UTC. This makes more sense since there is no "IsAdjustedToUTC" flag for thoses.


aloneguid:
Hi @Kuinox I've manually merged changes to master as PR was too old to safely merge. Closing this now. Thanks for all the work.

Kuinox:
No problem, thanks !

PR #696, opened 2025-12-28
Parellise columns writes.

Kuinox:
This PR allow to parallelise writes.  
It doesn't change the API, but the parallelisation control is limited due to that.  
The big constraint is that each WriteColumn must be started in order. As soon as it yield, we can start the next one.

But we can now easily introduce another API that allow to write columns in parallel.  

I moved a few things to static to avoid accidentaly modifying states concurrently.  

Each changes are in it's own commented commit, I'd recommand to read the changes commit by commit to follow it.

### DIFF
--- a/src/Parquet.Test/ParquetWriterTest.cs
+++ b/src/Parquet.Test/ParquetWriterTest.cs
@@ -6,6 +6,7 @@ using System.Collections.Generic;
 using System.Threading.Tasks;
 using Parquet.Schema;
 using System.Linq;
+using System.Threading;
 
 namespace Parquet.Test {
     public class ParquetWriterTest : TestBase {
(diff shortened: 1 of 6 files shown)


Kuinox:
Currently updating this to take into account changes on master.

Kuinox:
Should be good.  
There are still a lot of opportunities to reduce mutations on objects on the codebase.  
Having non mutable state, and reducing state, allows to make things multithreaded without big headaches.

aloneguid:
Sorry, I'll have to politely decline this feature. Writing this way may introduce unnecessary complexity. You can always write in parallel by dumping data to separate parquet files. I have no issues with code quality so don't take it personally please.

PR #712, opened 2026-03-01
feat: per-column dictionary encoding overrides and adaptive sampling

### LINKED ISSUE #639: Per column encoding
### Issue description

We would like to disable dictionary encoding for some of our double columns. I’m thinking about two ways of doing that:

1. Add a nullable `ParquetOptions` parameter `ParquetRowGroupWriter.WriteColumnAsync`, if the parameter is not `null` then it will be used instead of `_formatOptions`
2. Add a `Dictionary<string, Parquet.Meta.Encoding>` field to `ParquetOptions`, where the key is the column name. If a column name is not in the dictionary it will behave based on the values of the other fields of `ParquetOptions`. (This is basically equivalent to the column_encoding parameter in https://arrow.apache.org/docs/python/generated/pyarrow.parquet.ParquetWriter.html )

Both seem easy implement, any preference between or issues with the proposals?

meni-braun:
# Per-Column Dictionary Encoding Overrides and Adaptive Sampling

## In a Nutshell
For performance optimization, add two new `ParquetOptions` properties:
- `ColumnDictionaryEncodings`: per-column override dictionary (Dictionary<string, bool>?)
- `DictionaryEncodingSampleSize`: prefix sampling to skip high-cardinality columns (default 0, recommended: 1024)

This avoids expensive full-column `HashSet` scans on high-cardinality data, eliminating LOH allocations and reducing GC pressure in write-heavy workloads.

## Problem

When writing Parquet files with `UseDictionaryEncoding = true` (the default), the library attempts dictionary encoding on **every column in every row group** by building a `HashSet<string>` (or `Dictionary<T, int>`) over the **entire column data** before deciding whether to keep it. For high-cardinality columns (e.g., GUIDs, SHA-256 hashes, UPNs, URLs) this is pure waste:

1. **A `HashSet<string>` pre-allocated to `count` capacity** is created for each column - for a 100K-row row group with 20 string columns, that's 20 HashSets each sized for 100K entries. (In my production pipeline, I use 2,000,000-row row groups.)
2. **All values are hashed and inserted**, only to discover that `distinctCount / totalCount > threshold` and dictionary encoding is rejected.
3. These large collections are immediately discarded, generating **significant GC pressure**.

In .NET, any object larger than 85,000 bytes is allocated on the Large Object Heap (LOH). A `HashSet<string>` with 100K entries easily exceeds this - its internal `_buckets` array alone is ~400KB. When you have many string columns (as is typical in telemetry, log analytics, or SIEM pipelines), this pattern produces **megabytes of LOH allocations per row group**, all immediately discardable. The resulting Gen2/LOH GC collections can consume 15-20% of CPU in write-heavy services.

### Concrete Example

Consider a table with 20 string columns written in 100K-row row groups:

| Column Type | Count | Example Columns | Unique Ratio | Dictionary Useful? |
|-------------|-------|----------------|-------------|-------------------|
| High-cardinality | 12 (60%) | TransactionId, CorrelationId, FileHash, RequestUrl, TraceId | >95% | **No** - rejected after full scan |
| Medium-cardinality | 4 (20%) | TenantId, ServerIP, AppVersion | 10-50% | Yes |
| Low-cardinality | 4 (20%) | EventType, Status, Category | <5% | Yes |

**Current behavior:** All 20 columns get a full HashSet scan. The 12 high-cardinality columns each allocate a ~2.4 MB HashSet on the LOH, scan all 100K values, reject dictionary encoding, and discard the HashSet. That's ~29 MB of LOH waste **per row group**.

**With this PR:** The 12 high-cardinality columns are either skipped via explicit per-column override, or the adaptive sampler checks only the first 1,024 values, finds >80% unique, and bails out immediately - allocating a ~48KB HashSet (well under the 85KB LOH threshold) instead of 2.4 MB.

## Solution

This PR adds two complementary features to `ParquetOptions`:

### 1. Per-Column Dictionary Encoding Overrides (`ColumnDictionaryEncodings`)

A `Dictionary<string, bool>?` that lets callers explicitly enable or disable dictionary encoding for individual columns by path. Columns not listed fall back to the global `UseDictionaryEncoding` setting.

```csharp
var options = new ParquetOptions
{
    UseDictionaryEncoding = true,  // global default
    ColumnDictionaryEncodings = new Dictionary<string, bool>
    {
        ["FileHash"]       = false, // known high-cardinality, skip dictionary
        ["CorrelationId"] = false,
        ["RequestUrl"]    = false,
        ["EventType"]     = true,  // force dictionary even if global is false
    }
};
```

This is useful when the caller knows the schema ahead of time and can statically configure which columns benefit from dictionary encoding.

### 2. Adaptive Dictionary Encoding Sampling (`DictionaryEncodingSampleSize`)

An `int` property (default: `0` for not changing current behavior, although I think it's recommended: `1024` if it's OK to change that implementation detail) that enables prefix sampling before committing to a full dictionary scan. When greater than zero and the column has more values than the sample size:

1. Only the first `DictionaryEncodingSampleSize` values are inserted into a HashSet.
2. If the sample's unique ratio already exceeds `DictionaryEncodingThreshold`, dictionary encoding is **skipped immediately** - no full scan needed.
3. If the sample passes (low cardinality), the full scan proceeds as normal (the complete dictionary is needed for encoding).

```csharp
var options = new ParquetOptions
{
    UseDictionaryEncoding = true,
    DictionaryEncodingSampleSize = 1024,  // sample first 1024 values
};
// High-cardinality columns are automatically detected and skipped.
```

This is the "fire-and-forget" option: callers don't need to know which columns are high-cardinality. The sampler detects it automatically with negligible overhead.

### Priority Resolution

When both features are used together, the resolution order in `DataColumnWriter` is:

1. **Explicit per-column override** (`ColumnDictionaryEncodings[path]`) - highest priority, sampling is skipped
2. **Global setting** (`UseDictionaryEncoding`) + adaptive sampling (`DictionaryEncodingSampleSize`)

## Why Sampling Works

Dictionary encoding's value comes from **low-cardinality** columns (few distinct values relative to row count). The cardinality characteristics of a column are typically apparent within the first few hundred values:

- A `Status` column with 5 possible values will show <=5 unique values in any sample of 1,024.
- A `FileHash` column will show ~1,024 unique values in a sample of 1,024 (nearly 100% unique).

With sample size of 1,024 and threshold of 0.8:
- **True positive rate:** A column with <80% unique values in a 1,024 sample is extremely likely to be low-cardinality overall. The full scan will confirm and build the complete dictionary.
- **True negative rate:** A column with >80% unique values in a 1,024 sample is nearly certainly high-cardinality. Skipping the full scan is the correct decision.
- **False negative risk:** A column that appears high-cardinality in the sample but is actually low-cardinality overall. This is theoretically possible (e.g., the first 1,024 rows happen to all be unique, but the remaining 99K rows repeat). In practice, this is rare and the consequence is benign - the column simply uses plain encoding instead of dictionary encoding, which is slightly less compressed but still correct.

The sample HashSet for 1,024 string entries is approximately **48 KB** - well under the 85 KB LOH threshold. This is the key insight: sampling keeps the temporary allocation in Gen0/Gen1 territory, avoiding LOH pressure entirely.

## Benchmarks

We benchmarked three configurations on a realistic workload (20 string columns with mixed cardinality, 100K rows per row group):

### Single Row Group (100K rows)

```
BenchmarkDotNet v0.14.0, Windows 11
Intel Xeon Platinum 8370C CPU 2.80GHz, 1 CPU, 8 cores
.NET 8.0.24, X64 RyuJIT AVX-512F, Server GC

| Method                          | Mean       | Allocated  | Alloc Ratio |
|-------------------------------- |-----------:|-----------:|------------:|
| Dictionary ON (baseline)        | 271.7 ms   | 239.63 MB  | 1.00        |
| Dictionary OFF (blanket)        | 241.3 ms   | 190.47 MB  | 0.79        |
| Adaptive Sampling (this PR)     | 234.7 ms   | 223.39 MB  | 0.93        |
```

### 20 Row Groups (production-scale, 2M total rows)

```
| Method                          | Mean       | Allocated  | Alloc Ratio |
|-------------------------------- |-----------:|-----------:|------------:|
| Dictionary ON (baseline)        | 5.712 s    | 4403.98 MB | 1.00        |
| Dictionary OFF (blanket)        | 5.290 s    | 4616.73 MB | 1.05        |
| Adaptive Sampling (this PR)     | 4.780 s    | 4078.62 MB | 0.93        |
```

### Analysis

| Metric | Dictionary OFF vs Baseline | Adaptive vs Baseline |
|--------|---------------------------|---------------------|
| **Speed (1 RG)** | 11.2% faster | 13.6% faster |
| **Speed (20 RGs)** | 7.4% faster | 16.3% faster |
| **Memory (20 RGs)** | +4.8% more (no dict compression) | 7.4% less |
| **Dictionary preserved?** | No - all columns use plain encoding | Yes - low-cardinality columns still get dictionary |

The adaptive approach is the **fastest in both scenarios** - even faster than blanket disable - because it avoids both the full HashSet scan overhead AND preserves dictionary encoding for low-cardinality columns (which compress better, reducing downstream I/O). At production scale (20 RGs), it's **16% faster** and uses **7% less memory** than the baseline.

### LOH Impact (estimated from allocation profiles)

| Scenario | LOH Allocs per RG | Estimated GC Impact |
|----------|-------------------|-------------------|
| Dictionary ON (baseline) | ~29 MB (12 high-card x ~2.4 MB each) | 15-20% CPU in GC |
| Dictionary OFF | 0 MB | Minimal GC |
| Adaptive Sampling (1024) | ~0.6 MB (12 x ~48 KB, all under LOH threshold) | Minimal GC |

## Backward Compatibility

- **Fully backward compatible.** Both new properties have sensible defaults:
  - `ColumnDictionaryEncodings` defaults to `null` (no per-column overrides)
  - `DictionaryEncodingSampleSize` defaults to `0` (disabled - preserves existing full-scan behavior)
- To enable adaptive sampling, set `DictionaryEncodingSampleSize = 1024` (or any positive value). This is a pure optimization - it produces identical output for low-cardinality columns and skips dictionary encoding for high-cardinality columns that would have been rejected anyway after a full scan.
- `ColumnDictionaryEncodings` uses `IReadOnlyDictionary<string, bool>` to signal that it must not be modified while a writer is active.
- No breaking changes to any existing API surface.

## Related Issues

Closes #639

## Testing

Tests cover:
- Per-column overrides: enabling/disabling dictionary per column path, mixed with global setting
- Adaptive sampling: high-cardinality columns skipped, low-cardinality columns still get dictionary
- Edge cases: sample size larger than data, sample size of 0 (disabled), nested column paths
- Encoding validation: file size comparison proves dictionary presence/absence
- All 509 non-integration tests pass (9 pre-existing integration test failures unrelated to this change)

### DIFF
--- a/src/Parquet.Test/Encodings/DictionaryEncodingTest.cs
+++ b/src/Parquet.Test/Encodings/DictionaryEncodingTest.cs
@@ -1,4 +1,4 @@
-﻿using System;
+using System;
 using System.Collections.Generic;
 using System.IO;
 using System.Linq;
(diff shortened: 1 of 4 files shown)


aloneguid:
I really like this and you're right about parquet currently being really dumb trying to dictionary-encode everything. It's one of the major slowdowns for many scenarios.

I'm currently making breaking changes for v6 that will move low-level API to Span/Memory and will have a look at this, sorry for the delay.

aloneguid:
@meni-braun are you willing to adopt this to latest master? I know there might be a lot of changes.

meni-braun:
Sure, your wish is my command!
Will work on that.

meni-braun:
Hi,

I merged from latest master.
Whenever you have time, please give it a review.

Thanks!

aloneguid (approved):
Sounds good, thank you very much! This will be part of version 6 release.

PR #750, opened 2026-05-08
Update README.md

jamesonnetworks-clearcost:
Update class name to match what is in version 6.0.2, looks like it changed in code, but may not have changed in the docs. 

Thanks for all that you all do! Appreciate the code!


aloneguid (approved):
Thanks, appreciate it.

PR #759, opened 2026-05-21
Add back link to Parquet.FSharp

rob-earwaker:
The **Parquet.FSharp** link was accidentally removed in https://github.com/aloneguid/parquet-dotnet/pull/704 so I've added that back in.

I've also slightly reformatted the entire list so that the links only enclose the project titles and not the descriptions. I think this makes it a bit easier to identify projects and scan through the list - see before and after screenshots below. Happy to revert this reformatting change though, just let me know!

### Before reformatting

image

### After reformatting

image


aloneguid (approved):
Thank you.

Issue #776, opened 2026-08-03
[BUG]: Can't Read File With All Null Column since PyArrow v25


mukunku:
### Library Version

5.6.1

### .NET Runtime

.NET 10

### OS and Architecture

Windows (64-bit)

### How to reproduce?

Test file: [all_null_column_pyarrow_v25.zip](https://github.com/user-attachments/files/30682217/all_null_column_pyarrow_v25.zip)

It can be opened with https://parquetreader.com/

### Failing test

```csharp
public async Task AllNullColumnPyArrowV25() {
	using Stream s = OpenTestFile("all_null_column_pyarrow_v25.parquet");
	await using ParquetReader r = await ParquetReader.CreateAsync(s);
	using ParquetRowGroupReader groupReader = r.OpenRowGroupReader(0);
	DataField[] fs = r.Schema.GetDataFields();
	double?[] data = await ReadNullableValuesAsync<double>(groupReader, fs[0]);
	Assert.Equal(46, data.Length);
	Assert.All(data, d => Assert.Null(d));

	data = await ReadNullableValuesAsync<double>(groupReader, fs[1]);
	Assert.Equal(46, data.Length);
	Assert.All(data, d => Assert.Equal(0, d));
}
```

PR #777, opened 2026-08-03
Read Column With Null Data Page Header (#776)

mukunku:
# Summary

This is for #776. 

One of my users [reported](https://github.com/mukunku/ParquetViewer/issues/189) an issue when reading a file created with PyArrow v25 where all the values in the column are null. It seems the `DataPageHeader` is null for this file which causes the library to throw an exception when reading it. However other providers like parquetreader.com and DuckDB are able to open this file.

This PR is my clunky attempt to make this test file readable. I'm open to feedback or if you'd like to come up with your own solution 👍🏾


src/Parquet/File/DataColumnReader.cs
```diff
@@ -79,6 +79,7 @@ public async ValueTask ReadAsync<T>(ReadingColumn<T> rc, CancellationToken cance
                     break;
                 case PageType.DATA_PAGE:
                     await ReadDataPageV1Async(ph, rc, cancellationToken);
+                    allNullColumnProcessed = definedValuesCount == 0;
```
mukunku, line 82: Unfortunately `rc.ValuesRead` always returns 0 for this column because it has a MaxDefinitionLevel of 1. See:
```csharp
public int ValuesRead => Field.MaxDefinitionLevel > 0
    ? _definitionOffset
    : _definedDataCount;
```
So the `rc.MarkValuesRead(allValueCount);` call I added below isn't actually doing anything in this situation 🤷🏾‍♂️ 

This is my clunky attempt to get this to work. The loop will first read the dictionary page, then the data page and then will exit early.

This also doesn't make any changes to `ReadDataPageV2Async` so that's also an open question 😬

aloneguid:
@mukunku I see your point, but I don't think that's the correct fix. It seems like DataPage is null in c#, but not null in the metadata. This could be due to previous dictionary page declaring UncompressedSize as 0 and comparessed as 1, so there' some error in file offset calculations. Investigating.

aloneguid:
@mukunku PR #781 fixes this

mukunku:
Sounds good! Thanks for the fix 🙌🏾

Issue #726, opened 2026-04-09
Encryption support (footer key)


thunder7553:
### Issue description

I am currently using DuckDB.NET for reading parquet files, as this is the only library that I found that supports encryption. 

I see that there are two pull requests open that have something to do with encryption, and one issue that was closed due to inactivity. Since most tools do not support encryption, maybe this is not such a big topic in general for most users. 

Is there any chance that this will come to this library? I would rather have a managed library.

aloneguid:
That's something to be explored. Can you share a sample file and duck db query how exactly you are reading the encrypted files? Haven't ever seen those in practice.

thunder7553:
I was just following the docs on duckdb https://duckdb.org/docs/current/data/parquet/encryption . I have the same observations - most tools do not support encryption, because the underlaying libraries do not support it.  so I do not think it is wide spread (yet). From a GDPR perspective, unencrypted customer data on a temporary storage is still an issue, so I do like that. 

I will try to add a pull request with example files next week.

aloneguid:
Ah, that seems to be in the incubating stage as well. I think trying to read files encrypted with DuckDB will be a good first step. I can look at this after V6 is released if no one else will.

PR #782, opened 2026-08-11
Modular Encryption V2

kiloOhm:
Rebuilt from current master as suggested in #654 

## Summary

Adds Parquet Modular Encryption support to `parquet-dotnet` across the reader and writer paths.

## What changed

- Added public encryption and decryption configuration through `ParquetOptions`.
- Supports `AES_GCM_V1` and `AES_GCM_CTR_V1`.
- Supports encrypted footers and signed plaintext footers.
- Supports uniform footer-key encryption and per-column keys.
- Supports 128-, 192-, and 256-bit AES keys.
- Supports stored and externally supplied AAD prefixes.
- Added key resolution through direct keys or `IParquetKeyRetriever`.
- Encrypts and decrypts footers, column metadata, page headers, and page bodies.
- Allows reading available plaintext/projected columns without every column key.
- Added decryption-aware `ReadSchemaAsync` overloads.
- Fails closed on wrong keys, tampered modules, mismatched AAD prefixes, and unknown configured column paths.
- Explicitly rejects appending to encrypted files, which is not supported yet.

## Interoperability and testing

- Added parquet-java/parquet-mr interoperability coverage.
- Added external fixture coverage across:
  - both encryption algorithms
  - encrypted and plaintext footer modes
  - stored, supplied, and absent AAD prefixes
  - uniform and partial column encryption
  - 128- and 256-bit keys
  - key metadata configurations
- Replaced the committed 79 MB inspector JAR with pinned, reviewable Java source built by Maven in CI.
- Added DuckDB interop test #726

Validation completed with:

```text
dotnet test src/Parquet.slnx -c Release --no-restore
895 passed, 3 skipped, 0 failed


kiloOhm:
Just read the LLM rule. Closing

